PR #2, opened 2019-07-25
No preload, load/play on intersect, remove classes from imgs etc

ZaneHannanAU:


Soitora:
What's the point of the red dotted line if I may ask?

ZaneHannanAU:
Similar to https://1.1.1.1 (keep scrolling...)

PR #18, opened 2019-09-18
Use fetch polyfill instead of axios

arkon:


arkon:
Seems to not work, even with `whatwg-fetch` due to the nature of trying to do SSR. Going to abandon this, but at least it verified the new action works.

PR #19, opened 2019-09-19
Slightly enhance guide

Soitora:
Added reading progress as well as enhancing the install guide _slightly_.


arkon:
Reading progress seems overkill for the purposes of this site tbh

Soitora:
You really think so? I feel like it's a comfortable indicator on how far you've scrolled

Although it would be nice if it could be disabled on the main page/outside of wiki/guides.

arkon:
I think it's only useful for actual reading content like news articles, blogs, novels, etc. Otherwise, a regular old scrollbar is already sufficient.

Soitora:
I don't got much to say in this I suppose, removed the code for it.

src/README.md
```diff
@@ -11,7 +11,7 @@ features:
       details: MyAnimeList, AniList and Kitsu support.
     - title: Configurable
       details: Reader with multiple viewers, reading directions and other settings.
-footer: Apache Licensed | Copyright © 2015-present Javier Tomás
+footer: Designed with VuePress | MIT Licensed
```
arkon: Why? This is supposed to reflect info about Tachiyomi itself.

Soitora: I remember seeing the MPL license (which is why this repo has it too), my bad. Thought it was unrelated to Tachiyomi.

VuePress license does however require us to disclose the MIT license.

Soitora: How do we go about disclosing the MIT license then? Do we just change this repos license instead?

arkon: Why would an MIT license require that...?

arkon: FWIW, 1.x is licensed under MPL but 0.x is licensed under Apache.

Soitora: Alright so that's where I got the MPL license from.

Regarding the MIT license:
https://github.com/vuejs/vuepress/blob/master/LICENSE
> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

That's how I interpreted it considering the quote above. I'm no expert in copyrights however. Maybe it just means we can't remove it from the code hidden.. somewhere?

arkon: That's fine, it's already part of the Vuepress source code that's pulled in via NPM. You don't need to explicitly call out MIT licenses.

Soitora: Alright, I'll retract that commit and we should be ready to go.

PR #21, opened 2019-09-19
Fixed sidebar navigation for FAQ

Soitora:
Now that guides are separated too and they're both collapsible I felt very unnecessary to keep the landing page that essentially does nothing extraordinary.


Soitora:
I don't mind that but for now it's an annoying extra step

PR #27, opened 2019-09-20
Rework navigation

Soitora:
This is a big PR, I'm not sure I'm done but leaving it here for review under meanwhile.

It changes a lot of navigation/sidebars, adds a bit of new pages for guide/faq as well as offloads the nav/sidebar/plugin code to external files.

PR #30, opened 2019-09-23
Move navigation and plugins away from config.js (Split 1 of 3 - #27)

Soitora:
Moves `nav`, `sidebar` and `plugins` to outside of `config.js` as it was getting longer, will also be easier to manage.

This is the first PR after trying to split up #27.

One thing I'd like input on is the name "Regular" that is shown on the Nav-bar when expanding it, will try to come think of a better one but would look weirdly out of place if it didn't have a header to it when the coming commits will include one.

_**EDIT:** I built these three on top of each other, I'm guessing if something needs to be changed in this then I'll need to add a commit to it and it's descendants separately._


src/.vuepress/config/nav.js
```diff
@@ -0,0 +1,50 @@
+module.exports = [
+	{
+		text: 'Need help?',
+		items: [
+			{
+				text: 'Regular',
```
Soitora: Would prefer if we came up a better term for this, but will do for now.

src/.vuepress/config/nav.js
```diff
@@ -0,0 +1,50 @@
...
+					}
+				]
+			}
+		]
+	},
+	{
+		text: 'Guides',
+		items: [
+			{
+				text: 'Regular',
```
Soitora: Same goes here, but in this case it's slightly more fitting.

PR #32, opened 2019-09-23
Guide/FAQ overview page added (Split 3 of 3 - #27)

Soitora:
Re-added the Guide/FAQ overview pages as they are more necessary now and follow the changed structure.
These take advantage of the structure setup in #30 and new pages in #31 and is also dependent on them both.

This is the second PR after trying to split up #27.

Will need people to add onto and enhance the pages.

_**EDIT:** I built these three on top of each other, I'm guessing if something needs to be changed in the previous I'll need to add a commit to it and it's descendants separately._

_**EDIT 2:** I'm assuming you know how to only view files changed from the commit this version is dedicated too so it doesn't get lost in the sea of the previous commits 9206fe49d55bee08a66bff2b981d06bde7a570f5 and 8e3021280772c38014b163aab39827ffa8d7f1ed_


src/help/faq/README.md
```diff
@@ -0,0 +1,33 @@
+---
+title: Frequently Asked Questions
+lang: en-US
+---
+
+# Frequently Asked Questions
```
arkon: Shouldn't this help overview page live a level higher (e.g. `/help/README.md`)?

Soitora: Look line 38 in `sidebar.js`, same goes for line 5 (the path ends in a `/` so defaults to `README.md`). It can be a level higher but it's not needed and I felt it would be easier to have all related content in their respective folders.

PR #33, opened 2019-09-23
Update general.md

joseph619:
Update the General settings page with info

### DIFF
--- a/src/help/guides/general.md
+++ b/src/help/guides/general.md
@@ -15,9 +15,9 @@ Changes the app language. There are 35 languages to choose from, alognside the d
 By default, Tachiyomi is in light mode. To change the theme to dark mode, go to `Settings` -> `General` -> `Application theme`. There are four options:
 
 ### `Main theme` (light), 
-### `Dark theme`, 
+### `Dark theme` 
 ### `AMOLED theme`(pure black)
-### `Dark blue`. 
+### `Dark blue` 
 
 ## Library manga per row
 
--- a/src/help/guides/general.md
+++ b/src/help/guides/general.md
@@ -10,7 +10,7 @@ General settings page for Tachiyomi, help us improve through [here](https://gith
 
 Changes the app language. There are 35 languages to choose from, alognside the default system language. The instruction to change language can be found [here](../faq/application.html#how-do-i-change-the-app-language). 
 
-## Dark mode
+## Application theme
 
 By default, Tachiyomi is in light mode. To change the theme to dark mode, go to `Settings` -> `General` -> `Application theme`. There are four options, `Main theme` (light), `Dark theme`, `AMOLED theme` and `Dark blue`. AMOLED theme is pure black. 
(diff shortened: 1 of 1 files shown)


src/help/guides/general.md
```diff
@@ -4,34 +4,44 @@ lang: en-US
...
+General settings page for Tachiyomi, help us improve through [here](https://github.com/tachiyomiorg/website/edit/master/src/help/guides/general.md)!
 
 ## Language
-:construction: :construction_worker_man: W.I.P.
 
-## Application theme
-:construction: :construction_worker_man: W.I.P.
+Changes the app language. There are 35 languages to choose from, alognside the default system language. The instruction to change language can be found [here](../faq/application.html#how-do-i-change-the-app-language). 
+
+## Dark mode
```
Soitora: There are more themes than just Dark mode, this section should better represent themeing in general, although you can still emphasis the fact that it can be used to toggle dark mode. 

Maybe add sub-headers to this section `###` or `####` with each of the themes so it's searchable? (Although I don't know how far down the headers that search goes)

joseph619: I named it dark mode because that's what users would generally search for in the search bar. Especially those who're migrating from Manga Rock.

Since keywords are not implemented yet, I thought this would be a good temporary solution. But your choice.

joseph619: PR updated

joseph619: How do I add sub-header? Doesn't seem to be working.

joseph619: How do I change font size for sub-headers? It's too big

Soitora: Each `#` you add for the headers makes it smaller (increases in number), `####` (h4) should be suitable?

src/help/guides/general.md
```diff
@@ -4,34 +4,49 @@ lang: en-US
...
-:construction: :construction_worker_man: W.I.P.
+
+Changes the app language. There are 35 languages to choose from, alognside the default system language. The instruction to change language can be found [here](../faq/application.html#how-do-i-change-the-app-language). 
 
 ## Application theme
-:construction: :construction_worker_man: W.I.P.
+
+By default, Tachiyomi is in light mode. To change the theme to dark mode, go to `Settings` -> `General` -> `Application theme`. There are four options:
+
+### `Main theme` (light),
```
Soitora: Comma in the end of line 17?

joseph619: Hmm it's fixed, thanks for bringing it my attention

Soitora:
I'm going to accept #35 instead but great work nonetheless, your contributions are very much appreciated!

joseph619:
> I'm going to accept #35 instead but great work nonetheless, your contributions are very much appreciated!

May I know why you're using #35 instead of mine? It might be helpful for future.

joseph619:
Alright thank you

PR #38, opened 2019-09-24
Small styling tweaks

Soitora:
Changed logo size on front-page, added styling to the expanders as well as slimming down the scrollbars.

image


src/.vuepress/config/plugins.js
```diff
@@ -12,7 +12,7 @@ module.exports = [
 		'container',
 		{
 			type: 'expander',
-			before: info => `<details><summary>${info}</summary>\n`,
+			before: info => `<details class=\"expander_details\"><summary class=\"expander_summary\">${info}</summary>\n`,
```
arkon: I don't think you need the escape characters (`\`).

src/.vuepress/styles/index.scss
```diff
@@ -10,11 +10,71 @@
...
+
+.home .hero img {
+	max-height: 120px !important;
+}
+
+html {
+	overflow-y: scroll;
+}
+
+::-webkit-scrollbar {
```
arkon: Don't change the scrollbar styles, it's an anti-pattern. Leave "standard" things the way that users would expect them.

For example, I normally wouldn't even see the scrollbars on Android or macOS unless actually scrolling. This would force them to always show up.

src/.vuepress/styles/index.scss
```diff
@@ -10,11 +10,71 @@
...
+	background-color: #f3f5f7;
+	color: #2E84BF;
+	border-left: 0.5rem solid #2E84BF;
+}
+
+.expander_details strong {
+	-webkit-user-select: none;
+	-moz-user-select: none;
+	-ms-user-select: none;
+	user-select: none;
```
arkon: Why?

Soitora: When clicking to expand sometimes it introduced unnecessary and ugly accidental text highlights. This way it looks more like a real expanding element and not just a text that does things.

src/.vuepress/styles/index.scss
Soitora: This should not force them to always show up, at all actually as it's just modifying the current behaviour (except on Firefox). I don't mind removing it, I myself was a bit uncertain of using it.

src/.vuepress/config/plugins.js
Soitora: It didn't work the first time so I added escaping and it still didn't work and then it ("magically") just did, I forgot to check if it was related, I can do more testing on it.

Soitora: Okay I found out that it's just spoopy sometimes. No matter what I did the content on (let's say) `extensions.md` didn't change. But I found out that if you force refresh the file by adding and then removing something and saving in-between _then_ it will show the change. (ughh)

Issue #44, opened 2019-09-24
Redesign index page


Soitora:
I feel like the Index page could use a little touch up outside of the default theme layout, not really sure what though but it can be customized pretty heavily by editing the main theme.

Could be designed in a way that also allows for the download buttons to take place in a good manner (#43).

Issue #43, opened 2019-09-24
Parser for latest GitHub release


Soitora:
I'd like someone to make a parser in a way that allows us to have a button that downloads the latest `.apk` from GitHub. Using the `/release` URL isn't possible as it only allows direct-download if you specify the filename, but since we use the version in the APK filename that is a no-go.

The wanted tags is `tag_name` and `browser_download_url` from <https://api.github.com/repos/inorichi/tachiyomi/releases/latest>.

If you can, make a makeshift button for the front-page like the **Get Started** button that downloads APK directly.

alessandrojean:
I can help with that. Since it will require creating a custom layout, it will make #44 easier later.

arkon:
You pretty much just need to make a request for that URL, parse the JSON (e.g. with `JSON#parse`), then grab the fields from the resulting JS object.

alessandrojean:
That's what I am doing :)

PR #46, opened 2019-09-24
Add download link at index page

alessandrojean:
- Create a download icon in the home page that parses the GitHub releases API (#43).
- Create a custom theme to include the button in the home. 
  This way, the index page can be easily customized (#44).
- Add FontAwesome (with the Vue wrapper), used for the download icon at index and extensions. 
  The way it's added, it only adds the used icons to the build, not importing all of them.

Maybe it's interesting in the future create a store using Vuex to use the [advantages of the SSR](https://ssr.vuejs.org/guide/data.html) and generate the index page HTML already with the latest download link.

### DIFF
--- a/src/.vuepress/theme/components/Home.vue
+++ b/src/.vuepress/theme/components/Home.vue
@@ -51,12 +51,12 @@
 
 		<Content class="theme-default-content custom"/>
 
-		<div
+		<footer
 			class="footer"
 			v-if="data.footer"
 		>
 			{{ data.footer }}
-		</div>
+		</footer>
 	</main>
 </template>
 
@@ -65,7 +65,7 @@ import NavLink from '@parent-theme/components/NavLink.vue';
 
 import axios from 'axios';
 
-const RELEASE_URL = 'https://api.github.com/repos/inorichi/tachiyomi/releases/latest'
+const RELEASE_URL = 'https://api.github.com/repos/inorichi/tachiyomi/releases/latest';
 
 export default {
 	components: { NavLink },
--- a/src/.vuepress/theme/components/Home.vue
+++ b/src/.vuepress/theme/components/Home.vue
@@ -1,0 +1,254 @@
+<template>
+	<main class="home" aria-labelledby="main-title">
+		<header class="hero">
+			<img
+				v-if="data.heroImage"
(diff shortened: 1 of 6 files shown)


src/.vuepress/theme/components/Home.vue
```diff
@@ -0,0 +1,256 @@
+<template>
+	<main class="home" aria-labelledby="main-title">
+		<header class="hero">
+			<img
+				v-if="data.heroImage"
+				:src="$withBase(data.heroImage)"
+				:alt="data.heroAlt || 'hero'"
+			>
+
+			<h1 v-if="data.heroText !== null" id="main-title">{{ data.heroText || $title || 'Hello' }}</h1>
```
arkon: Is the fallback really necessary? Same goes for some other stuff like the description. I haven't really worked with Vue much, but it'd be nice if you could just mark them as required props (attributes?) for the component.

src/.vuepress/theme/components/Home.vue
```diff
@@ -0,0 +1,256 @@
...
+				:key="index"
+			>
+				<h2>{{ feature.title }}</h2>
+				<p>{{ feature.details }}</p>
+			</div>
+		</div>
+
+		<Content class="theme-default-content custom"/>
+
+		<div
```
arkon: Could use `<footer>`

src/.vuepress/theme/components/Home.vue
```diff
@@ -0,0 +1,256 @@
...
+		</div>
+	</main>
+</template>
+
+<script>
+import NavLink from '@parent-theme/components/NavLink.vue';
+
+import axios from 'axios';
+
+const RELEASE_URL = 'https://api.github.com/repos/inorichi/tachiyomi/releases/latest'
```
arkon: Missing semicolon.

src/.vuepress/theme/components/Home.vue
alessandrojean: The component does not use properties, instead it reads the content from the YAML frontmatter in the Markdown. So I think it's not possible to make them required, so they needs the fallback. It's the way that the default theme [does](https://github.com/vuejs/vuepress/blob/master/packages/%40vuepress/theme-default/components/Home.vue).

PR #48, opened 2019-09-25
Update extensions.md

joseph619:
Added Manga Rock error solution

### DIFF
--- a/src/help/faq/extensions.md
+++ b/src/help/faq/extensions.md
@@ -195,6 +195,10 @@ Some extensions contain multiple sources, look at the list below to find out whi
 
 Try closing (force-quit) and reopening Tachiyomi.
 
+### Error: Loader not implemented
+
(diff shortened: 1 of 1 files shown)


src/help/faq/extensions.md
```diff
@@ -238,6 +242,10 @@ There is a tool to migrate from MangaRock to Tachiyomi courtesy of a member of o
 
 Note that you must download the 64-bit version of Java or it will run out of memory. Access the program at [waicool20/mr2tachiyomi](https://github.com/waicool20/mr2tachiyomi).
 
+### Error: Value <url> at 0 of type java.lang.String cannot be converted to JSONObject
```
Soitora: Is it really worth adding more MR entries though considering it will soon be unusable?

joseph619: I am hearing that they won't be closing the site as soon as promised, so I guess we can keep it around for now. We can remove it once they truly shut it down.

Soitora: Alright that's fine.

PR #55, opened 2019-09-29
Added how to see latest manga in a source

joseph619:


src/help/faq/extensions.md
```diff
@@ -84,6 +84,10 @@ Be aware that Batoto is deprecated so you will need to migrate from that source
 10. WieManga
 :::
 
+### How do I read `Latest` manga?
```
Soitora: I actually didn't understand this at all, because it's called `Catalogues` and not `Galleries`.
Also I think this section should be placed as a subsection somewhere on guides instead of FAQ, if you want it here until such entry is made at least fix the Galleries thing and maybe rephrase title as:
> `### How do I see the latest manga from a source?`

joseph619: You're right, I messed up. Eh still calls it Galleries instead of Catalogues. I will fix it asap.

Maybe we should also create sections for 

My Library
Library updates
Recently read
Catalogues
Extensions
Download queue

Soitora: I don't mind that, had planned to do so all along, wait until tomorrow before doing anything in regard to that though.

Soitora:
Change the section title according to feedback and we're good.

src/help/faq/extensions.md
joseph619: Everything done

joseph619: I guess I could move this to Catalogues section once it's written.

PR #70, opened 2019-10-02
Reword/Enhance all of FAQ

Soitora:
Spent a lot of time touching up FAQ making it look more uniform and not... cluttered.

PR #71, opened 2019-10-03
Fix the commit history for PR #70

Flat:


Soitora:
Is this supposed to be a replacement for #70 or merged after #70?

PR #76, opened 2019-10-04
Create LegacyMaterialIcon.vue

### LINKED ISSUE #73: Fix MDI-Glasses component
I introduced a very simple MDI-Glasses component in #70, it's a legacy icon from Material Design (current is Material Icons), I basically colored an SVG instead.

If possible it would be good to be able to refer to old MD icons like the MI ones added by @GHOSTBEAR in #64.

GHOSTBEAR:
Closes #73 

Adds ability to use Material Design icons 

Uses http://materialdesignicons.com/ for font

## Code example
`<MaterialIcon icon-name="mdi-glasses" legacy />`


arkon:
Could it be added to the other component but with a legacy flag as a prop?

GHOSTBEAR:
@arkon I have added legacy flag as you requested

arkon:
Could you also replace/remove that glasses component that was added in #70? 😅

PR #77, opened 2019-10-04
Replace videos and their components

Soitora:
Rerecorded the videos with a specific template, used FFMPEG to add transparent padding then the frame around it. Also replaced the carousel with the new one (#75) in one of the video sections, but not all, that comes in a later commit. 

Overall the size of the video library should have halved as well.


arkon:
In the future, keep independent changes in separate PRs -- e.g. re-recorded videos can be in one PR, code changes in another.

Soitora:
@arkon Alright, I thought this was specific enough but I got you

PR #83, opened 2019-10-06
Configure automatic linting

arkon:
Will need to investigate running it automatically as either a pre-commit hook or a GitHub action.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,22 @@
+{
+	"env": {
+		"browser": true,
+		"es6": true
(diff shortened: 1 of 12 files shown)


arkon:
Interestingly the build worked locally... I'll need to look into it more.

Soitora:
How's this going? I saw an error related to Font Awesome, which we're depracating right? Not sure what else failed

arkon:
I still don't understand why it failed on deployment but not locally, but it seems to run now...

PR #84, opened 2019-10-07
Add name prop to MaterialIcon

GHOSTBEAR:
Adds ability to style MaterialIcon

## Code Example
``` html
<MaterialIcon icon-name="extension"/>
<MaterialIcon name="download" icon-name="extension"/>
<MaterialIcon icon-name="mdi-glasses" name="download" legacy/>
<MaterialIcon name="notDownload" icon-name="report"/>
```

``` css
.download {
  font-size: 64px;
  color: red;
} 

.notDownload {
  font-size: 48px;
  color: yellow;
}
```

## Styling only the icon
``` css
.download .material-icons {
  color: red;
} 

.download .material-legacy-icons {
  color: green;
} 
```

### DIFF
--- a/src/.vuepress/components/MaterialIcon.vue
+++ b/src/.vuepress/components/MaterialIcon.vue
@@ -1,6 +1,6 @@
 <template>
-	<div class="icon-holder">
-		<i v-if="legacy" class="legacy-material-icons mdi" v-bind:class="iconName"></i>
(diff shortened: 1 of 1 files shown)


arkon:
Shouldn't it be called `class`?

Soitora:
The other components uses it too, and I think it's good with distinguishment when it's not technically putting a straight class on it

Issue #26, opened 2019-09-20
Add keywords such as 'Auto Backup' for search returns


joseph619:
Add keywords for important settings/features/FAQ which will refer to a relevant section, such as 'Auto Backup' which will take the user to Auto Backup section.

PR #85, opened 2019-10-09
Switch to Algolia DocSearch

Soitora:
Switches from the regular search to the much more powerful Algolia one, not sure if it's ready yet or not but I'd prefer a better way to style it rather than doing `!important` in the `.scss` file.

PR'd so you can preview the build.


joseph619:
Nice! This should also take care of #26

src/.vuepress/styles/_variables.scss
```diff
@@ -0,0 +1,46 @@
+// SEARCHBOX
```
arkon: Are these used by Algolia or something?

src/.vuepress/styles/index.scss
```diff
@@ -108,3 +108,29 @@
...
+}
+
+.algolia-autocomplete .algolia-docsearch-suggestion--text {
+	font-size: 0.8rem !important;
+	color: #808080 !important;
+}
+
+.algolia-autocomplete .algolia-docsearch-suggestion--highlight {
+	color: #2E84BF !important;
+}
```
arkon: Could nest these in `.algolia-autocomplete`.

You could _potentially_ avoid `!important` if you make the selectors more specific somehow.

src/.vuepress/styles/index.scss
```diff
@@ -108,3 +108,29 @@
...
+
+.algolia-search-wrapper .algolia-autocomplete .algolia-docsearch-suggestion .algolia-docsearch-suggestion--subcategory-column {
+	background: #FFFFFF !important;
+}
+
+.algolia-search-wrapper .algolia-autocomplete .algolia-docsearch-suggestion .algolia-docsearch-suggestion--subcategory-column-text {
+	color: #A1A1A1 !important;
```
arkon: Lots of hex codes happening in here. I wonder if we should have variables for these somewhere.

src/.vuepress/config.js
```diff
@@ -10,6 +10,11 @@ module.exports = {
 		repo: 'inorichi/tachiyomi',
 		docsRepo: 'tachiyomiorg/website',
 		docsDir: 'src',
+		algolia: {
+			apiKey: 'fc1c45b5a3835e1882cbbf0751dfe705',
+			indexName: 'tachiyomi',
+			debug: true
```
arkon: Is this required?

src/.vuepress/styles/_variables.scss
Soitora: You were supposed to be able to edit it but I think you need to eject the Algolia parts from the them and then edit, which I'm not sure how to do.

src/.vuepress/styles/index.scss
Soitora: Just saw it as a temporary solution, best would be doing it with the intended file but I was unsuccessful in setting that up.

src/.vuepress/styles/index.scss
Soitora: Definitely, when looking at SCSS vs Stylus the other day I was reminded of its variable feature (sure, CSS has :root but the SCSS variables are better.).

src/.vuepress/config.js
Soitora: Index Name and API Key? Yes.
Debug? No.

Debug just prevents the pop-up from disappearing when for example opening DevTools, which makes styling easier.

Production should not have that enabled.
(There's `.gitignore`, but can you setup so that you can ignore a line?)

arkon: The comment was for L16 (`debug: true`). I realize GitHub doesn't make it obvious which lines a comment is referring to very well...

Could maybe do this: https://stackoverflow.com/a/49310302

src/.vuepress/styles/_variables.scss
arkon: Yeah: https://community.algolia.com/docsearch/styling.html#advanced-styling

We can just remove this if it's not being used.

Soitora:
Overhauled everything, I can't avoid the `!important` styles yet because the component is wonky (and I don't know how to eject) (https://github.com/vuejs/vuepress/issues/331)

**There are some issues which is mainly due to our setup and not Algolia itself:**
- Sometimes when searching a header is empty, I guess because it's searching top-level and can't find an above header.
- Headers with badge next to them gets included as pure text
- Text with icon next to them gets the icon name included as pure text

arkon:
This seems relevant: https://community.algolia.com/docsearch/tips.html

Soitora:
That's good documentation, can maybe look into it today after I visit the ER

Soitora:
Got Algolia to reindex the page (it's supposed to be automatic but I guess not touching it for ~200 days made it go idle), also changed from using a stylesheet to replace styles I just modified the ejected component instead.

src/.vuepress/config.js
```diff
@@ -11,6 +11,11 @@ module.exports = {
 		docsRepo: 'tachiyomiorg/website',
 		docsDir: 'src',
 		logo: '/assets/media/logo.svg',
+		algolia: {
+			apiKey: 'fc1c45b5a3835e1882cbbf0751dfe705',
```
arkon, line 15: Is this supposed to be public?

src/.vuepress/theme/components/Navbar.vue
```diff
@@ -0,0 +1,140 @@
...
+    const handleLinksWrapWidth = () => {
+      if (document.documentElement.clientWidth < MOBILE_DESKTOP_BREAKPOINT) {
+        this.linksWrapMaxWidth = null
+      } else {
+        this.linksWrapMaxWidth = this.$el.offsetWidth - NAVBAR_VERTICAL_PADDING
+          - (this.$refs.siteName && this.$refs.siteName.offsetWidth || 0)
+      }
+    }
+    handleLinksWrapWidth()
+    window.addEventListener('resize', handleLinksWrapWidth, false)
```
arkon, line 83: Could also use https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia , which is more performant

Soitora: Not sure how I'd add that, this file is just ejected VuePress file, not something I wrote on my own, kind of how `Home.vue` is just an edited default file and not custom written

src/.vuepress/config.js
Soitora: Yes, it's an alright key to be "public", it's implemented exactly according to their spec

PR #89, opened 2019-10-14
Update home and extension icons to MaterialDesignIcons

alessandrojean:
Removes previous FontAwesome icons and replaces them with MaterialDesignIcons for consistency.

### DIFF
--- a/src/.vuepress/theme/components/Home.vue
+++ b/src/.vuepress/theme/components/Home.vue
@@ -24,7 +24,7 @@
 					title="Download latest release"
 					download
 				>
-					<font-awesome-icon icon="download" />
+					<MaterialIcon icon-name="get_app"/>
 					<span>Download {{ tagName || 'vX.X.X' }}</span>
 				</a>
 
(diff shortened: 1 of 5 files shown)


Soitora:
Maybe try getting the download button content more centered? I know m.vlad was very strict so I had to spend a while making my button in **Getting Started** "perfect".
Or maybe reuse those buttons to `Home.vue`?

alessandrojean:
Sure, I can do it :)

src/.vuepress/theme/components/Home.vue
```diff
@@ -24,8 +24,10 @@
 					title="Download latest release"
 					download
 				>
-					<font-awesome-icon icon="download" />
-					<span>Download {{ tagName || 'vX.X.X' }}</span>
+					<div class="icon-holder">
+						<MaterialIcon icon-name="get_app"/>
+						<span>Download {{ tagName || 'vX.X.X' }}</span>
```
Soitora: Maybe do this as it looks like less is changing on load, or have it set to latest release (`0.8.6`) as the updates are few, and if a new one comes it will be replace anyways.
```suggestion
						<span>Download {{ tagName || 'v0.0.0' }}</span>
```

Soitora:
Right now the initial load looks like this:
image
Do you think you can fix it or shall I try it when my time frees up?

alessandrojean:
That's strange. I can try to fix it.

alessandrojean:
I tried, but I don't know what is causing this behavior. :(

As you suggested, can you try to fix it when you have some time?

PR #92, opened 2019-11-01
Info about Multi-source extensions

joseph619:


Soitora:
I can't really approve this as is.

Instead, maybe replace **Mangakakalot** text with **MangaBox** and then add a **Mangakaklot** section under **KissManga** with the FAQ in question placed.

The factoid about multi-source I don't feel is needed.

joseph619:
I see. Unfortunately I don't have the time to make changes, so please add info about Mangakakalot moving in a new FAQ section yourself.

PR #94, opened 2019-11-14
Makes download link point to github release if github is slow, fails …

ZaneHannanAU:
Ensures download is available if js is blocked, some part is down, or latency is extreme.

Just kinda adds a default state.


Soitora:
Good thinking, already had this implemented for the Getting Started buttons, but they have a regular `download` string and not the `:download="browserDownloadUrl ? '' : null"`. What does that add exactly in functionality @ZaneHannanAU ?

ZaneHannanAU:
download can be empty or null (empty is download normal, null is no download).
Basically makes it a normal link.

PR #105, opened 2020-01-08
Change index download buttons

Soitora:
Change the **Stable** download button on the index/front-page to the dual download buttons featured on the "Getting started" page.

image

### DIFF
--- a/src/.vuepress/styles/index.scss
+++ b/src/.vuepress/styles/index.scss
@@ -115,3 +115,11 @@
 	-ms-user-select: none;
 	user-select: none;
 }
+
+a.nav-link.action-button.secondary {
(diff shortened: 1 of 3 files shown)


arkon:
Wondering if we should use this as an opportunity to start pointing users to dev instead...

joseph619:
We should, Stable is getting too old, ideally we want to promote something that has active development. Stable never had a feature freeze this long, it's abandoned at this point.

Soitora:
I feel that Dev has always been stable enough, never felt that Stable has really been "needed". I personally would prefer people to be on Dev because it allows for easier documentation and fixes.

arkon:
I guess we can do that separately.

PR #109, opened 2020-01-20
Removed the word root and added info about local

joseph619:
Removed the word root (which caused confusion among readers), and added info that local has to be created manually first.

### DIFF
--- a/src/help/guides/reading-local-manga.md
+++ b/src/help/guides/reading-local-manga.md
@@ -5,9 +5,9 @@ lang: en-US
 
 # Reading local manga
 
-You have to first create a folder named `local` in Tachiyomi folder, which is located in the phone's `Internal storage` or external SD card. Then place your manga inside the folder `Tachiyomi/local`, usually `/sdcard` and `/storage/18F5-2C11` (with different values).
+You have to first create a folder named `local` in Tachiyomi folder, which is located in the phone's `Internal storage` or external SD card. Then place your manga inside `Tachiyomi/local` folder.
 
-After that, they will be shown in the catalogue as a new source.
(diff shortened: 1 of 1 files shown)


src/help/guides/reading-local-manga.md
```diff
@@ -5,7 +5,7 @@ lang: en-US
 
 # Reading local manga
 
-You have to place your manga inside the folder `Tachiyomi/local` located in the root of your internal storage or external SD card, usually `/sdcard` and `/storage/18F5-2C11` (with different values).
+You have to first create a folder named `local` in Tachiyomi folder, which is located in the phone's `Internal storage` or external SD card. Then place your manga inside the folder `Tachiyomi/local`, usually `/sdcard` and `/storage/18F5-2C11` (with different values).
```
arkon: Did you mean to keep the ending?

joseph619: Thanks for replying, what do you mean by 'the ending' here?

arkon: The ", usually `/sdcard` and `/storage/18F5-2C11` (with different values)" part. It sounded like you were trying to reword it to be less confusing, but now it just doesn't make sense.

joseph619:
Oh, that part wasn't written by me. I had kept that part untouched, should I rewrite it as well?

arkon:
Yes, the sentence doesn't actually flow in its current state.

src/help/guides/reading-local-manga.md
joseph619: Oh, that part wasn't written by me. I left it untouched, should I rewrite that as well?

joseph619:
Ok I will edit that too

PR #111, opened 2020-02-04
[Experimentation] Add navbar logo

Soitora:
Experimenting with some stuff offered by VuePress.

I'd like input on wether or not to add a navbar logo at all, and if so, does it need further tweaks?

image


arkon:
Maybe use a PNG instead?

https://caniuse.com/#feat=webp

Soitora:
Switched to SVG, any device too old to render it can't run the app anyways.

PR #112, opened 2020-02-04
Switch to Stylus and cleanup related code

Soitora:
- Switched all `CSS`/`SCSS`/`SASS` to `Stylus` _(after careful consideration and testing I find this the most compatible with the work we're doing.)_
- Removed redundant classes _(such as `no-select`)_
- Remove `index.styl` import as it was causing duplicates _(it imports by default, importing again showed duplicate everything in DevTools)_
- Filled `palette.styl` with the default settings
- Switched Mobile responsive breakpoint to `767px`
- Reverted the border color on tip containers _(it was really looking too bland with just blue-ish colors everywhere, I can put this in a seperate PR if wanted)_


src/.vuepress/components/DownloadButtons.vue
```diff
@@ -42,77 +42,64 @@
...
 
-<style scoped lang="scss">
-* {
-	font-family: "Open Sans", Roboto, -apple-system, BlinkMacSystemFont,
-		"Segoe UI", Oxygen, Ubuntu, Cantarell, "Fira Sans", "Droid Sans",
-		"Helvetica Neue", sans-serif;
-}
+<style scoped lang="stylus">
+*
+    font-family 'Open Sans', Roboto, -apple-system, BlinkMacSystemFont, 'Segoe UI', Oxygen, Ubuntu, Cantarell, 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif
```
arkon: Aren't we using tabs in this project?

src/.vuepress/styles/index.styl
```diff
@@ -1,37 +1,123 @@
-.guide
-	margin 1rem 0
-	padding .1rem 1.5rem
-	border-radius 0.4rem
-	background-color #f0f4f8
-	.title
-		font-weight bold
+// Custom rules
+user-select()
```
arkon: Can you use this in `DownloadButtons.vue`?

src/.vuepress/styles/palette.styl
```diff
@@ -1 +1,22 @@
...
+$badgeWarningColor = darken(#ffe564, 35%)
+$badgeErrorColor = #DA5961
+$badgeVersionColor = #000000
+$containerBackground = #f0f4f8
+
+// Layout
+$navbarHeight = 3.6rem
+$sidebarWidth = 20rem
+$contentWidth = 740px
+$homePageWidth = 960px
```
arkon: Should probably convert these into `rem`s too.

src/.vuepress/components/DownloadButtons.vue
Soitora: Shit, VS Code fails me yet again despite following the editorconfig

src/.vuepress/styles/index.styl
Soitora: I actually don't know, but I can try and see if referencing it works, would make code even slimmer

PR #117, opened 2020-03-02
Track download modal clicks

Soitora:
_Attempts_ to add ("Debugging the `ga();` tag was difficult in localhost") tags for tracking download events.

I haven't been able to try this as GA don't at all want to function in localhost. This is the correct implementation I'm pretty sure though.

If this doesn't work then I read that it might require switching from the GA Plugin to a standard `.js` script with GA Global Tag support.

Also updates dependencies alongside it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,21 +13,33 @@
 				"@babel/highlight": "^7.0.0"
 			}
 		},
-		"@babel/core": {
-			"version": "7.7.7",
-			"resolved": "https://registry.npmjs.org/@babel/core/-/core-7.7.7.tgz",
-			"integrity": "sha512-jlSjuj/7z138NLZALxVgrx13AOtqip42ATZP7+kYl53GvDV6+4dCek1mVUo8z8c8Xnw/mx2q3d9HWh3griuesQ==",
+		"@babel/compat-data": {
+			"version": "7.8.6",
+			"resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.8.6.tgz",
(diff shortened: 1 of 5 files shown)


Soitora:
Okay so, it was as I assumed, `ga` is not available from the plugin alone.

arkon:
Probably want to track the getting started guide buttons too.

arkon:
You can check if `ga` is available on the global scope on the live website. The linter will complain about it even if it is available.

Soitora:
You're right, I overlooked the **Getting started** buttons.
Running `window.ga` in console yields `ƒ (a){J(1);Z.D.apply(Z,[arguments])}`, yields `undefined` in localhost

Soitora:
You think it could work by appending `window.` in front of `ga`?
That's what was suggested for the linter here: https://stackoverflow.com/a/44196695

arkon:
Worth a shot. You should be able to run the linter locally too.

Soitora:
Running the linter locally with the exact same line that was shown in Netlify changed *a ton* of files.
Redid the download buttons, you will probably find stuff to comment on there.

Preferably will have Index and Getting Started share a big chunk of the code, they aren't right now as they function differently.

arkon:
> Running the linter locally with the exact same line that was shown in Netlify changed a ton of files.

Weird. Could also be due to bumping up the Prettier config package, but I'm not really concerned about that.

---

It might be more useful to have different metrics for the different places. I haven't worked with GA much, but I think being able to tell the difference would be useful -- unless it already groups the metrics based on URL?

Soitora:
I think it can, or might group the metrics based on the URL as I've seen a lot of that before, in this case I'll just go with your suggestion and seperate it so you're sure to be seeing right.

PR #125, opened 2020-03-05
Fix text overflow on index

Soitora:
Fix text overflow on index like @joseph619 experienced when using 120% font scaling on phone browser.

Solved by instead of having a set `height: 5em` I changed it to `min-height: 4em` so that it want to stay a uniform size but can be increased.

Prevents this:
image


Soitora:
When I first created it I recall adding a height to it because otherwise the images would get uneven from the text size difference. I'll double-check soon if that really was the case

PR #130, opened 2020-03-20
Add Spreadsheet to Extensions Page

### LINKED ISSUE #127: Add Spreadsheet to Extension List
It might be nice to add a reference to the spreadsheet on the extension list. https://tachiyomi.org/extensions/

happywillow0:
Closes #127 

Also, Cloudflare Captcha is a http `403` rather than a `503`. 
503 is normally seen with the I'm under attack / automatic browser check which errors out as `Failed to bypass cloudflare`.

### DIFF
--- a/src/extensions/README.md
+++ b/src/extensions/README.md
@@ -7,4 +7,10 @@ lang: en-US
 
 Here is a list of all available extensions to download inside the app.
 
+::: guide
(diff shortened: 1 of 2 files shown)


Soitora:
I really would prefer a built-in "show on website" solution, but for now I'll accept this in wait for something better.

PR #143, opened 2020-04-24
Update GA event names

Soitora:
Since GA event names preferably has to be updated due to the Dev -> Preview change, I thought I'd take the time to make it more fitting for reading data.

### DIFF
--- a/src/.vuepress/components/DownloadButtons.vue
+++ b/src/.vuepress/components/DownloadButtons.vue
@@ -1,10 +1,10 @@
 <template>
 	<div class="downloadContainer">
 		<button class="downloadStableButton" @click="downloadStable">
-			Stable
+			{{ downloadStableLabel }}
 		</button>
-		<button class="downloadDevButton" @click="downloadDev">
-			Dev
+		<button class="downloadPreviewButton" @click="downloadPreview">
+			{{ downloadPreviewLabel }}
 		</button>
 	</div>
 </template>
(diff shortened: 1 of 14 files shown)


arkon:
Couldn't you add a dimension to indicate what page it's coming from and use the same metric name across them?

Soitora:
I can see if I can grab the page name from Vuepress and inject that so it doesn't need to be hardcoded.

Soitora:
Okay so it's probably not exactly what you wanted, but, it didn't make sense for 100% static Home / Fork to have variable metric names. Although I did change the `DownloadButtons.vue` so that you can change both the metric name and label name, it will also show which page it's situatated on. Is this okay?

arkon:
I meant something like this: https://developers.google.com/analytics/devguides/collection/analyticsjs/custom-dims-mets

So you could in theory be able to track an event called `Download stable`, but also have a dimension called `Page` with values like `Homepage` or `Forks`. That way you could aggregate the numbers from GA more easily (or at least I assume you can).

Soitora:
Okay so, I set it up, but I'm not sure how it will actually work soo that needs testing, but since GA debug doesn't work in local enviroment but it does in Netlify, I hope it'll give me something satisfactory

Soitora:
Clicking "Stable" button in "Getting started" yields (amongst others):
> **hitType** (event)
> **eventCategory** (Button)
> **eventAction** (Click)
> **eventLabel** (User download)
> **metric1** (Tachiyomi)
> **dimension1** (Getting started)

Doing the same in Index:
> **hitType** (event)
> **eventCategory** (Button)
> **eventAction** (Click)
> **eventLabel** (User download)
> **metric1** (Tachiyomi Preview)
> **dimension1** (Home)

This goes the same for example Neko, except I added the dimension to page title, which I shouldn't have, I should have set it as a static "Forks", need to change that real quick.

src/.vuepress/components/DownloadButtons.vue
```diff
@@ -84,13 +122,25 @@ export default {
...
 				}
 			});
-			window.location.assign("https://tachiyomi.kanade.eu/latest");
+			window.location.assign(
+				this.$props.downloadPreviewUrl || PREVIEW_URL
+			);
+			window.ga(
+				"set",
+				"metric1",
+				this.$props.downloadStableTag
```
arkon, line 131: Preview tag?

Soitora: Shit, this is why we have reviews :)
I'll fix it after my food coma is over

Soitora:
Going to do some quick testing with Netlify and see how it tracks before merge

PR #144, opened 2020-04-26
Add DummyKiss for FAQ download

Soitora:
As requested by Support staff, adding linkable DummyKiss.

Although GitHub certainly isn't a file host per-se, a file or two shouldn't be a big issue.


arkon:
It already complains that we're using GH Pages to host files. We should considering throwing these into a separate branch or repo, maybe?

Soitora:
A seperate repo might be too much for a few files.
Branch maybe, could that still be easily tied to the `tachiyomi.org` domain, or do you mean to use the `raw.github` links?

arkon:
> or do you mean to use the `raw.github` links?

Yeah, that.

Soitora:
So a branch called `files` with a `README.md`, `LICENSE` and the files, at top level, or do you have a better idea?

arkon:
Don't really need the README or LICENSE files. Just make sure it's an orphan branch, similar to `repo` in the extensions repo.

Soitora:
Alright, and is this just for DummyRock/DummyKiss files-ish or do you intend for the video files too?

arkon:
We can start off with the DummyKiss files first.

Soitora:
Done, should maybe protect so that branches can't be created by anyone, seeing many of the PRs done here by others being branches here instead of branches on their own forks

PR #146, opened 2020-04-27
Remove "menus" guides

Soitora:
I thought it was a decent idea at first to include, but in all honesty it feels redundant and a waste of effort.

We had like two linked items to the page, for those small stuff it's better to include in FAQ or not at all.


arkon:
If menus need a guide, we should focus on coming up with ideas to improve them instead. :)

PR #154, opened 2020-04-30
Update MaterialIcon.vue

GHOSTBEAR:
Removes Material Design Icons because it doesn't work, and didn't find anything that uses it.
Also, added the ability to style iconOnly flag.

Before (Upper left material icon, upper right iconOnly material icon, lower part MDI):
image
After:
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2686,11 +2686,6 @@
 				"to-fast-properties": "^2.0.0"
 			}
 		},
-		"@mdi/font": {
-			"version": "5.1.45",
-			"resolved": "https://registry.npmjs.org/@mdi/font/-/font-5.1.45.tgz",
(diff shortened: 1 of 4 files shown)


Soitora:
I think @arkon has changed all icons in the app to use the newer Material Design no?

GHOSTBEAR:
Have seen commits about updating icon in the app, so maybe all icons is Material Icons

arkon:
I'm not super clear on which font provides what icons, so...

GHOSTBEAR:
Material Icons the font which is left is from https://material.io/resources/icons/. While Material Design Icons font is from https://materialdesignicons.com/ which just provides alot of icons.

Soitora:
Looks good, there's no icons on the website using legacy anymore 👍

PR #157, opened 2020-05-01
[0.9.0] Replace screenshots on homepage

arkon:


Soitora:
Maybe compress them a little, their size was increased 300%

arkon:
Done.

PR #161, opened 2020-05-04
Updated Sources problems as per 0.9.0

joseph619:


joseph619:
There are two Browse, one is the bottom tab, the other is similar to LATEST. I added item only for the bottom tab one which has an icon.

I will capitalize the Webviews as requsted.

src/help/guides/source-problems.md
```diff
@@ -46,19 +46,22 @@ In this case it is a problem with the site.
...
 If there is no CAPTCHA on the webpage, initiate one by tapping a manga title, if possible. Once a CAPTCHA is shown, solve it. After that, use the aforementioned back arrow on the bar at the top of your screen to go back to your manga. Please do not use the back button on your phone as it does not work as expected.
 
 ::: tip
-If WebView doesn't work for you, please update the **Android System WebView** app through [Play Store](https://play.google.com/store/apps/details?id=com.google.android.WebView), then go to [Developer Options](https://developer.android.com/studio/debug/dev-options) in your phone's settings and set WebView Implementation to Android System WebView.
+If WebView doesn't work for you, please update the **Android System WebView** app through [Play Store](https://play.google.com/store/apps/details?id=com.google.android.WebView), then go to [Developer Options](https://developer.android.com/studio/debug/dev-options) in your phone's settings and set **WebView Implementation** to Android System WebView. If Android System Webview doesn't work, set **Google Chrome** as default in Webview Implementation.
```
Soitora, line 64:
```suggestion
If WebView doesn't work for you, please update the **Android System WebView** app through [Play Store](https://play.google.com/store/apps/details?id=com.google.android.WebView), then go to [Developer Options](https://developer.android.com/studio/debug/dev-options) in your phone's settings and set **WebView Implementation** to Android System WebView. If Android System Webview doesn't work, set **Google Chrome** as default in WebView Implementation.
```
How I mean by capitalizing WebView

joseph619: Oh that. Yeah it's fixed.

PR #164, opened 2020-05-04
Updated Categories as per 0.9.0

joseph619:
All sections also need updated videos (which have been removed in this edit). Bottom two have been kept which also need to be replaced.

### DIFF
--- a/src/help/guides/categories.md
+++ b/src/help/guides/categories.md
@@ -9,57 +9,48 @@ lang: en-US
 
 ## Create a Category
 
-To make a category, go to <Navigation item="old_library"/>, then click on the <Navigation item="old_overflow"/> on the top right, then click **Edit categories** and then press the button with a plus sign **(`+`)**.
+To make a category follow the steps below.
 
-::: expander <strong>Click to reveal video guide!</strong>
-<figure class="centered">
-	<video muted loop controls :poster="$withBase('/assets/media/video-guide-category-create.png')" height="500" controlslist="nodownload noremoteplayback" preload="none" loading="lazy" crossorigin="use-credentials">
-		<source :src="$withBase('/assets/media/video-guide-category-create.webm')" type="video/webm"/>
-	</video>
-</figure>
-:::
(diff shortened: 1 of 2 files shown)


Soitora:
I believe this error is due to you referencing a navigation item that does not exist

src/help/guides/categories.md
```diff
@@ -9,57 +9,48 @@ lang: en-US
...
-<figure class="centered">
-	<video muted loop controls :poster="$withBase('/assets/media/video-guide-category-edit.png')" height="500" controlslist="nodownload noremoteplayback" preload="none" loading="lazy" crossorigin="use-credentials">
-		<source :src="$withBase('/assets/media/video-guide-category-edit.webm')" type="video/webm"/>
-	</video>
-</figure>
-:::
+1. Go to <Navigation item="more"/> → <Navigation item="settings"/> → <Navigation item="library"/>.
+1. Tap on the **Edit categories**.
+1. Long press a category.
+1. Click <Navigation item="edit"/> on top right.
```
Soitora, line 26:
```suggestion
1. Click **EDIT** on top right.
```
Probalby it's this that does not exist.

joseph619: Hmm I am not sure what the name is for the pencil icon, upon Long pressing it was labelled as Edit on the app.

Soitora: What I have added for the <Navigation> element you can find here: https://github.com/tachiyomiorg/website/blob/master/src/.vuepress/components/Navigation.vue

joseph619: I added it under `/* Actions */`, should I change it to under `/* Misc */`?

Soitora: No Actions is probably better

joseph619: Hmm, so what do I do now?

Soitora: > Hmm, so what do I do now?

Rebase your branch on top of the current upstream/master, probably

PR #178, opened 2020-05-09
Remove about page descriptions

arkon:
The page isn't really useful.


Soitora:
Agreed, was more useful before when user could search for how to turn on auto updates.

PR #193, opened 2020-05-13
Fix weird clipping in Help section

GHOSTBEAR:
Fixes weird clipping on the Discord card when screen is narrow
Current: 
image
After:
image


arkon:
I'd rather it actually pad/align to a grid properly. Not sure if @Soitora has thoughts about it.

Soitora:
A real grid might be better for sure, if someone wants to refactor my code then I'd appreciate that. I'm not great at making grids/responsive layouts

GHOSTBEAR:
image
I change it to display as a grid instead and all cards height follows to the tallest one

Soitora:
I didn't notice until now but, can you sneak in a hover fix for the border moving the content?
Most likely because it's 2px on hover and 1px on static

GHOSTBEAR:
@Soitora it should be fixed now

Soitora:
Perfect

PR #194, opened 2020-05-13
Rework Home page from scratch

Soitora:
Old `Home.vue` was during a long period of time occasionally added on to. This rework basically pulls a completely fresh and new version of the component and attempts to "cleanly" rewrite in the new data.

Better download popup with added a11y changes to them all, allowing for non-mouse enviroments which was previously unsupported.

Also added icons to the new buttons.


src/.vuepress/theme/components/Home.vue
```diff
@@ -73,211 +109,363 @@ export default {
...
+			max-width: 100%
 		h1
 			font-size 3rem
-		h1
-		.description
-		.action
-			margin 1.8rem auto
+		h1, .description, .action
+			margin 1rem auto
 		.description
```
arkon, line 297: You'll need to make sure these align properly. They sometimes have 2 or 3 lines, so the screenshots below are shifted.

src/.vuepress/theme/components/Home.vue
```diff
@@ -73,211 +109,363 @@ export default {
...
+		margin-right 25% !important
+		margin-bottom 8px !important
+	&__cancelButton
+		background none !important
+		color $textColor !important
+		font-size 0.8rem !important
+		padding 2px !important
+		padding-top 4px !important
+		border-top 1px solid darken($borderColor, 10%) !important
+		border-radius 6px !important
```
arkon, line 403: Shouldn't this be 0? There's a weird curve to the top border otherwise.

src/.vuepress/theme/components/Home.vue
Soitora: Yeah I'll look into it

src/.vuepress/theme/components/Home.vue
Soitora: Right, forgot that. It was added in last minute when I added the a11y focus effects, didn't notice the border line bending

PR #195, opened 2020-05-14
Update homepage text to match app features

specterflare:
The current homepage has a couple of inaccuracies and omissions:
1) The Kissmanga extension is no longer available or supported.
2) Support for the tracker Bangumi has been added.
3) 'Extensions' and 'Tracking' are both nouns, but 'Configurable' is an adjective, so it feels a bit off compared to the others. I changed it to 'Customization".


Soitora:
I'll wait with this PR until #194 pulled as it will not be fully compatible without changes

PR #196, opened 2020-05-14
Functionality: Themify navigation icons

Soitora:
Pretty bold change.

I had to use Vue Themify as I could not figure out how to import components from another file.
Here it's just setting an attribute so that works as planned but Material Icons is an entire component, calling it just prints it literally.

Icon list: https://z3by.github.io/vue-themify-icons/icons/

### DIFF
--- a/src/.vuepress/theme/components/NavLinks.vue
+++ b/src/.vuepress/theme/components/NavLinks.vue
@@ -1,0 +1,172 @@
+<template>
+  <nav
+    v-if="userLinks.length || repoLink"
+    class="nav-links"
+  >
+    <!-- user links -->
+    <div
+      v-for="item in userLinks"
+      :key="item.link"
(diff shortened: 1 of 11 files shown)


src/.vuepress/theme/components/NavLinks.vue
```diff
@@ -0,0 +1,172 @@
+<template>
```
arkon, line 1: Indentations seems off in this file. Are all of these ejected files or something?

Soitora: They are

arkon:
Was I supposed to review this again? I think I substituted "Experimentation" with "WIP" when I looked at it before.

Soitora:
Uhh I really don't remember, it has been a while ^^
Might need to rebase it and look it over again

Issue #202, opened 2020-05-16
Side navigation get overlay onto content


GHOSTBEAR:
This occurs on more narrow screens
And is prominent on FAQ, Guides, Fork, Contribution, and Extensions

Images of issue:
![image](https://user-images.githubusercontent.com/6576096/82119247-207d2500-977d-11ea-9234-60f5ea2c8861.png)

![image](https://user-images.githubusercontent.com/6576096/82119215-fcb9df00-977c-11ea-8d80-51f951ca88d2.png)

Soitora:
Might be due to the padding override I did for Help.vue (if that was the one).

PR #203, opened 2020-05-16
Remove PWA

Soitora:
As I had hoped would not happen, people fall for the "Install app" message that comes up when PWA is active.


Soitora:
Honestly not sure why this fails, local doesn't give any sort of error...

arkon:
I cleared the cache and it still hits the same issue. Maybe try just removing your `node_modules` and `package-lock.json` and rebuilding it?

specterflare:
I don't know if it's okay for me to comment here, but this PR breaks #202 (which was fixed by #204) again by shifting the content back to the left.

Soitora:
> I don't know if it's okay for me to comment here, but this PR re-breaks #202 (which was fixed by #204) by shifting the content back to the left.

This should be incorrect, this PR is built prior to the sidebar fix, thus it will not display those changes in the Netlify preview. Nothing changed in this PR could break that.

src/.vuepress/config.js
```diff
@@ -7,17 +7,11 @@ module.exports = {
 
 	// prettier-ignore
 	head: [
-		// PWA headers
+		// Embed headers
 		['link', { rel: 'icon', href: '/favicon.ico' }],
 		['link', { rel: 'manifest', href: '/manifest.json' }],
```
arkon, line 12: Doesn't this need to be removed too?

Soitora: Manifest.json?
That's for the embeds and android shortcuts, it's stuff that I would use regardless of PWA

Soitora: Not sure if Discord has changed but last time I fixed embeds for websites it needed some stuff from Manifest, but that was more if you wanted to specify more embed information like big image, small image, widge image, thumbnail or photo etc

arkon: Ah, I see. Not very familiar with that stuff.

PR #215, opened 2020-05-21
Deleted outdated video guides

joseph619:


GHOSTBEAR:
Source files could also be deleted. They are located under `src -> .vuepress -> public -> assets -> media`

joseph619:
Done

Soitora:
Kind of closed by #222

PR #227, opened 2020-06-07
Remove service worker

GHOSTBEAR:
This removes the deprecated service worker if any user has it

Tested this by loading in a dummy service worker
image
And then on refresh, it is removed
image


src/.vuepress/public/scripts/remove_service_worker.js
```diff
@@ -0,0 +1,11 @@
+console.log("Checking for service worker...");
+navigator.serviceWorker.getRegistrations().then(function (registrations) {
```
arkon, line 2: Shouldn't you check if `serviceWorker` is available?

GHOSTBEAR: It should, yes

PR #229, opened 2020-06-09
Add back service worker

arkon:
Need to test this with Netlify.

Hoping that by reinstating the SW, it'll pull in the newer content in the cache. I still left the script that's supposed to unregister it though, which should theoretically unregister the old existing service worker?


arkon:
It does seem to unregister and register again, so it _should_ at least solve the existing stale page issue.

Not entirely sure what the path forward is though. Maybe just a dummy `service-worker.js`?

Soitora:
As someone on Vuepress issues suggested, what happens if you add now "serviceWorker: false" as a plugin setting? The person said it should make existing service workers unregister

arkon:
It's still registered.

Soitora:
Yeah, it's unfortunate they didn't originally include something to clear the service worker if it was found not to be installed

Guess we'll keep both of them for a while

PR #236, opened 2020-06-13
Put the 'Q' back in 'FAQ'

specterflare:
This has actually been bugging me for a while

PR #237, opened 2020-06-13
Added FAQ about app not Uninstalling

joseph619:


src/help/faq/README.md
```diff
@@ -86,6 +86,10 @@ You can change language by going to <Navigation item="more"/> → <Navigation it
 :::
 
 ### Miscellaneous
+
+#### I can't uninstall Tachiyomi
+On **Android 10**, a bug was introduced which prevented users from uninstalling the app if the phone had a MicroSD card. To fix the issue, please remove the MicroSD card from phone first, then uninstall Tachiyomi.
```
GHOSTBEAR, line 91:
```suggestion
On **Android 10**, a bug was introduced which prevented users from uninstalling the app if the phone had a MicroSD card. To fix the issue, please remove the MicroSD card from the phone first, then uninstall Tachiyomi.
```
You are just missing a `the` before `phone` in `To fix the issue, please remove the MicroSD card from phone first`

specterflare:
Could you turn the header into a question, like "Why can't I uninstall Tachiyomi?"? I'm working on #236 and I think it would be best if the site was consistent.

src/help/faq/README.md
joseph619: Done

joseph619:
> Could you turn the header into a question, like "Why can't I uninstall Tachiyomi?"? I'm working on #236 and I think it would be best if the site was consistent.

Done

Issue #10, opened 2019-09-18
Consolidate information


arkon:
The website should ideally act as the single source of truth for FAQ and other similar information.

- [x] https://github.com/inorichi/tachiyomi/wiki
- [x] https://github.com/tachiyomiorg/website/wiki/FAQ
- [x] Discord FAQ

alessandrojean:
I can help with that if you guys need it, such as starting to complement the FAQ.

arkon:
Sure, just make sure to post what you'll be working on so others don't end up doing the same thing

alessandrojean:
Okay. I will start to complement the website FAQ with the other questions from this repo's wiki. Since they are categorized by sections (Application and Extensions) and their subcategories, I will see if I can split the FAQ in a file for the app questions and another one for the extension questions.

arkon:
That actually related to another thing suggested on Discord, which I've logged as #11.

Not sure if you're on Discord and looking at the discussion in the #dev-website channel.

alessandrojean:
The link on the issue it's exactly what I have in mind, separate the questions into categories. I will start only separating the existing questions for now.

specterflare:
So I don't know if you want to close this issue or not, but the checklist is complete now.

PR #258, opened 2020-07-11
Make use of "Help Center" more consistent

specterflare:


specterflare:
Unrelated, but you probably want to unpin #10, as it is now closed.

PR #266, opened 2020-07-18
Adds download page

Soitora:
As per the Discord discussions.
Something small and easy, might be worked on more in the future but the goal is to get it indexed as the top result on Google.

image

### DIFF
--- a/src/download/README.md
+++ b/src/download/README.md
@@ -13,3 +13,7 @@ Latest release: **<VersionTag/>**
 :::
 
 <DownloadButtons downloadStableTag="Tachiyomi" downloadPreviewTag="Tachiyomi Preview"/>
+
+::: guide Release notes
+<ReleaseNotes/>
+:::
--- a/src/download/README.md
+++ b/src/download/README.md
@@ -1,0 +1,15 @@
+---
+title: Download
+description: Download links for the latest versions of Tachiyomi.
+lang: en-US
+---
+
+# Download
+
+Download the latest version of **Tachiyomi** for your Android device.
+
+::: aside-guide
+Latest release: **<VersionTag/>**
+:::
+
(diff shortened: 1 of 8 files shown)


arkon:
Maybe you could also just modify the meta tags to improve the SEO?

Soitora:
Meta tags for this page, or the main home page?

I think Frontmatter allows some meta data tags

arkon:
The main home page.

Soitora:
Aside from [adding meta tags through frontmatter](https://v1.vuepress.vuejs.org/guide/frontmatter.html#meta), SEO optimization has never been something I've been decent at. Not sure how to tweak it good enough for people to stop using other download sites

GHOSTBEAR:
Something that I think could be added to this page is release notes for stable releases. But that could be addressed in another PR

Soitora:
> Something that I think could be added to this page is release notes for stable releases. But that could be addressed in another PR

I am testing that right now, but got rate limited 😞

Soitora:
Now has changelog (with correct markdown) too thanks to @Flat 



Anything else that should/could be tweaked?

src/download/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+# Download
+
+Download the latest version of **Tachiyomi** for your Android device.
+
+::: aside-guide
+Latest release: **<VersionTag/>**
+:::
+
+<DownloadButtons downloadStableTag="Tachiyomi" downloadPreviewTag="Tachiyomi Preview"/>
```
arkon, line 15: I don't remember, but does this distinguish what page the click metrics are being collected from?

Soitora: These in perticular is the name that gets recorded, so switched out for forks etc

It doesn't record which page

I'd use Google Tag Manager or something but it was too complex last I tried so silently gave up

arkon:
I'd still consider just adding some tags to the main page too, like "download", "Android app", "official", "manga", etc.

PR #271, opened 2020-07-28
Store data in Vuex

GHOSTBEAR:
This stores data in Vuex as the title suggests. Currently, it only stores the JSON object from the latest stable releases. But could be extended for other types of data, that could be shared by other components in the project


src/.vuepress/store/index.js
```diff
@@ -0,0 +1,52 @@
...
+Vue.use(Vuex);
+export default new Vuex.Store({
+	state: {
+		stable: {
+			updated: null,
+			data: null,
+		},
+	},
+	mutations: {
+		stable(state, stable) {
```
arkon, line 20: I haven't used Vuex before, but are names typically this vague? Maybe something like `setStableReleaseData`?

src/.vuepress/store/index.js
```diff
@@ -0,0 +1,52 @@
...
+			// eslint-disable-next-line no-param-reassign
+			state.stable = stable;
+		},
+	},
+	actions: {
+		async stable({ commit }) {
```
arkon, line 26: Same for this. `getStableReleaseData`?

src/.vuepress/store/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+			const now = new Date().getTime();
+
+			if (updated != null && now - updated <= 60 * 60 * 24 * 1000) {
+				return this.state.stable;
+			}
+
+			if (call == null) {
+				call = axios.get(RELEASE_URL);
+				promise = call.then(({ data }) => {
```
arkon, line 37: You may want to handle errors too, e.g. if GitHub is down.

src/.vuepress/store/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+			const now = new Date().getTime();
+
+			if (updated != null && now - updated <= 60 * 60 * 24 * 1000) {
+				return this.state.stable;
+			}
+
+			if (call == null) {
+				call = axios.get(RELEASE_URL);
+				promise = call
```
arkon, line 38: Shouldn't you be wrapping this whole call in a new promise and then calling resolve/reject?

src/.vuepress/components/DownloadButtons.vue
```diff
@@ -10,10 +10,7 @@
 </template>
 
 <script>
-import axios from "axios";
-
-const RELEASE_URL =
-	"https://api.github.com/repos/inorichi/tachiyomi/releases/latest";
+const RELEASE_URL = "https://github.com/inorichi/tachiyomi/releases/latest";
```
arkon, line 13: You'll probably need to rebase on top of #273 after that's merged.

src/.vuepress/store/index.js
GHOSTBEAR: I guess you mean the `promise` variable. What it does is basically just making any calls to the function wait for network call to be finished and then returning a promise that is resolved. And the reason for that is because I only want to make one network call if the data need to be updated

arkon: I mean that it should be:

```js
promise = new Promise((resolve, reject) => {
  call
    .then(({data} => {
        // ...
        resolve();
    }
    .catch((err) => {
      // ...
      reject(err);
    });
});

try {
  await promise;
} catch (err) {
  // log it or something?
}
call = null;
promise = null;
return this.state.stable;
```

GHOSTBEAR: Sure. But I already catch the network error and log it, and `this.state.stable` which gets returned will always have a Boolean saying if there was an error to the component/view. I wouldn't see the reason to use `reject` in `promise` because it will always update the data accordingly.

src/.vuepress/theme/components/Home.vue
```diff
@@ -79,17 +61,16 @@
...
 
 <script>
-import axios from "axios";
-
 import CloudDownloadIcon from "vue-material-design-icons/CloudDownload.vue";
 import BookOpenVariantIcon from "vue-material-design-icons/BookOpenVariant.vue";
-
-const RELEASE_URL =
-	"https://api.github.com/repos/inorichi/tachiyomi/releases/latest";
+<<<<<<< HEAD
```
arkon, line 66: Your branch seems borked.

GHOSTBEAR: More like I messed up my git rebase

Soitora:
You can merge it when you've checked so it's all correct, non-borked. ^^

GHOSTBEAR:
Think the only thing broken right now is the commit history

Soitora:
Only needs to be slightly fixed for #277, commit history not super important as we squash commit anyway. When you've fixed the conflicts and it looks good then you're good to merge

GHOSTBEAR:
After reading some more about `git rebase` it seems that I missed `git push --force` which is why I messed up the git history. I will open a new pull request to get a cleaner PR

PR #272, opened 2020-07-28
Removing Mangalife and Mangasee for being slow

joseph619:


joseph619:
MangaHub is pretty close so yeah

GHOSTBEAR:
Okay, MangaHub is already mentioned as " All around good source", so it would feel repetitive to add as "high-quality scans" too

PR #273, opened 2020-07-28
Create global constants file

Soitora:
Instead of having all files use their own `const` of the re-used assets such as URLs for download, put them in a global `constants.js` file and import to each file when needed.

### DIFF
--- a/src/.vuepress/components/DownloadButtons.vue
+++ b/src/.vuepress/components/DownloadButtons.vue
@@ -11,11 +11,8 @@
 
 <script>
 import axios from "axios";
-
-const RELEASE_URL =
-	"https://api.github.com/repos/inorichi/tachiyomi/releases/latest";
-
-const PREVIEW_URL = "https://tachiyomi.kanade.eu/latest";
+// eslint-disable-next-line prettier/prettier
+import { githubLatestApi, githubLatestRelease, kanadeLatest } from "../constants";
 
 export default {
 	props: {
@@ -53,7 +50,7 @@ export default {
 	},
 
 	async mounted() {
-		const { data } = await axios.get(RELEASE_URL);
+		const { data } = await axios.get(githubLatestApi);
 		// Maybe eventually some release has more than the apk in assets.
 		const apkAsset = data.assets.find((a) => a.name.includes(".apk"));
 		// Set the values.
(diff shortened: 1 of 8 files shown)


src/.vuepress/components/DownloadButtons.vue
```diff
@@ -11,11 +11,8 @@
 
 <script>
 import axios from "axios";
-
-const RELEASE_URL =
-	"https://api.github.com/repos/inorichi/tachiyomi/releases/latest";
-
-const PREVIEW_URL = "https://tachiyomi.kanade.eu/latest";
+// eslint-disable-next-line prettier/prettier
```
arkon, line 14: Why?

src/.vuepress/constants.js
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable prettier/prettier */
+export const githubExtensionJson = "https://raw.githubusercontent.com/inorichi/tachiyomi-extensions/repo/index.json"
```
arkon, line 2: Shouldn't these be `SNAKE_CASE`?

Soitora: Is there a specific reason we don't want to follow `camelCase` like we do for the other stuff?

src/.vuepress/components/DownloadButtons.vue
Soitora: Some lines are like 2 characters too long which makes it new-line, looks really awkward when only some are multi-line

Don't know how to change config for that behavior

src/.vuepress/constants.js
arkon: Just seems like the norm for constants.

src/.vuepress/components/DownloadButtons.vue
arkon: I'm pretty firmly in the camp of just leaving it as is, but I don't really maintain this project so it's up to you.

I think you can adjust `printWidth` for Prettier though

src/.vuepress/constants.js
Soitora: I'm fine with that

src/.vuepress/components/DownloadButtons.vue
Soitora: Alright no problem, adjusted Prettier to another recommended `printWidth` setting

PR #277, opened 2020-07-31
Enhance download page

### LINKED ISSUE #267: Better incorporate Downloads page
I think it would make more sense if the Download button on the home page redirected to the downloads page, instead of splitting up downloads into two locations. I'd make a PR myself if I knew how to do this lol.

Soitora:
## TL;DR
- Front-page download button now navigates you to the /download/ page.
- Add fallbacks for some components if GitHub fails to reply (i.e 403)
- Adds in the description a relative date presenter of when the latest release happened. _(Uses moment package, on hover a user is presented a readable exact time)_
- Tweak the buttons to include icons too _(Planning a future PR where I make a `Button.vue` component to clean their code and make it not as messy as it is now)_
- Changed "Release notes" to "Changelog"
- Removed the V-prefix from the release tag

## Preview
image

Resolves #267


src/.vuepress/components/ChangeLog.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+	<div class="guide changeLog">
+		<p class="title">Changelog</p>
```
arkon, line 3: You could word it as "What's new" similar to the app.

Soitora: I can

PR #278, opened 2020-08-01
Update categories for new app update

joseph619:


joseph619:
I updated the patch

src/.vuepress/components/Navigation.vue
```diff
@@ -61,15 +61,17 @@ export default {
 				/* More */
 				downloaded_only: { text: "Downloaded only", icon: "cloud_off" },
 				download_queue: { text: "Download queue", icon: "get_app" },
-				source_migration: { text: "Source migration", icon: "compare_arrows" },
+				categories: { text: "Categories", icon: "label" }
```
arkon: Build fails because you missed a comma.

joseph619: Yep fixed

arkon:
The build still fails.

You need to update the usages of these keys if you're changing them.

joseph619:
Hmm, I guess I will put them back in for now then

joseph619:
Should be fine now, build is no longer failing

src/.vuepress/components/Navigation.vue
```diff
@@ -83,6 +83,10 @@ export default {
 				share: { text: "Share", icon: "share" },
 				webview: { text: "WebView", icon: "public" },
 				/* Browse */
+				button_latest: { text: "LATEST" },
+				button_browse: { text: "BROWSE" },
+				tab_sources: { text: "SOURCES" },
+				tab_extensions: { text: "EXTENSIONS" },
```
GHOSTBEAR: If you planned to remove these they are still in use, and you will need to remove the components that are using them to have a successful build. You can use the GitHub repository search function to find the uses if you don't use VSCode or something

joseph619: I would remove them later, just want to update the page with new content for now

arkon:
I suspect the reference that the branch is diffing against might be messed up or something. Will need to clean stuff up at some point.

PR #281, opened 2020-08-02
Add instructions to create backup

Seishirou101:


AbdullahM0hamed:
Backups are an illusion.

PR #286, opened 2020-08-04
Remove source recs permanently

joseph619:
Also gave instruction to users why no recs


AbdullahM0hamed:
Recs are an illusion.

PR #297, opened 2020-08-15
Enhance slightly for SEO

Soitora:
Added `robots.txt`.
Very slightly tweaked front-page contrast colors, shouldn't be noticable but passes WCAG 2.1 AA standards.
Added link rels for the buttons (tbh not sure exactly if this helps)
Added alt text to the feature images
Added APK as a keyword for Download page (Although I'm pretty sure keywords don't affect anything)

### DIFF
--- a/src/.vuepress/config/plugins.js
+++ b/src/.vuepress/config/plugins.js
@@ -60,7 +60,6 @@ module.exports = [
 		"vuepress-plugin-robots",
 		{
 			host: "https://tachiyomi.org",
-			disallowAll: false,
 			allowAll: true,
 			sitemap: "/sitemap.xml",
 		},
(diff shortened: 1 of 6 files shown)


src/.vuepress/config/plugins.js
```diff
@@ -56,4 +56,13 @@ module.exports = [
 			hostname: "https://tachiyomi.org",
 		},
 	],
+	[
+		"vuepress-plugin-robots",
+		{
+			host: "https://tachiyomi.org",
+			disallowAll: false,
```
GHOSTBEAR, line 63: `disallowAll` is false by default so it not really necessary here

Soitora: I suppose that's true

PR #332, opened 2020-08-30
Quick "rewrite"

Soitora:
Okay so, I've felt that VuePress has been buggy in some areas, maybe it was due to our stuff not working as expected? Wrong. Apparently after starting over from a clean slate it's mostly just that VuePress *is* buggy.

Anyway, when starting over I decided to carefully set the package file using `npm init`, filling out some extra stuff including sorting the packages.

## Changelist
- As per Vuepress guide, `.eslintrc.json` is now `.eslintrc.js`.
- `package.json` has been reconstructed from a clean slate. (Might need Netlify `node_packages` rebuild).
- Some redundant and unused eslint stuff in `package.json` has been removed.
- `material-design-icons` was replaced with the 4 years newer `material-design-icons-iconfont`.
- As with changing MD icons, I've cleared some stuff in the `MaterialIcon.vue` file, the SCSS in it were necessary to cooperate.
- `plugins.js` now is sorted by the order they're presented in `package.json`, exception being `vuepress-plugin-container`
- `PWA` was turned off, it has been three months now since we had the PWA issues. As I saw Vuepress staff comment on months ago, turning it off should also clear it, but just in case I've still kept the pwa-remove script.
- `fonts.styl` was unused and thus it was removed
- `index.styl` now styles `body` and `html` too for the background color, this fixes background color on some pages where content is short.
- Download buttons in `Getting started` were removed in favor of redirecting users to the Download page.


src/.vuepress/components/MaterialIcon.vue
```diff
@@ -20,14 +20,27 @@ export default {
...
+$material-design-icons-font-directory-path: "~material-design-icons-iconfont/dist/fonts/";
+@import "~material-design-icons-iconfont/src/material-design-icons";
+
+.my-face {
+	@include material-icon("face");
+}
+
+.my-face:before {
+	content: material-icons-content("face");
+}
```
GHOSTBEAR, line 41:
```suggestion
.material-icon-face {
	@include material-icon("face");
}

.material-icon-face:before {
	content: material-icons-content("face");
}
```
Can the name of the CSS selector be more descriptive?

Soitora: Forgotten code, it's not necessary for our uses

GHOSTBEAR (approved):

PR #348, opened 2020-09-15
Switch last updated to relative timestamp

Soitora:
Switch from American timestamp to relative date marker.
E.g: "2 months ago" instead of "9/14/2020, 8:40:00 PM"

## Before:
image

## After:
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,8 @@
       "version": "1.0.0",
       "license": "MPL-2.0",
       "dependencies": {
+        "@mr-hope/vuepress-plugin-last-update": "^0.7.2",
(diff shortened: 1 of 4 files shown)


Soitora:
Love it when it works locally but won't build on Netlify, I'll look into it when I got free time

Soitora:
So basically "sitemap" was dependant on "last-updated", but it could not read the timestamp since it was now modified. What MR-Hope has done is basically modify the default plugin so that when the timestamp is tweaked for "last-updated" it will instead fetch the timestamp directly from Git.

ghostbear (approved):

PR #363, opened 2020-09-27
Containers overhaul

Soitora:
Tweaked custom containers to look more neutral but at the same time give them a sense of personality.
The three custom icons (tip, warning, danger) were provided by **TF#1071** on the Discord server.
## Example
| New | Original |
| --- | -------- |
| image| image |
| image | image |


ghostbear (commented):
Could the background hue be different for each container for example danger has a slightly red background and warning a slightly yellow background? The container for tips could remain the same.
To further illustrate this: 
image

They don't need to be intensely red or yellow just a slight hint

Soitora:
Sure, absolutely

Soitora:
I made the title built-in again, by remembering you can do `:empty` to specify stuff without content to hide it.
I did give blue a slight tint to make it more visible.

Also slightly lowered right padding and gave all containers a bottom margin.

Here's an image of how the objects look with the tweaks:
image

ghostbear (approved):

PR #372, opened 2020-10-01
Unify location for assets.. again..

Soitora:
Realized that keeping them seperate was a bad idea, especially for newcommers that want to contribute.

PR #373, opened 2020-10-01
Local guide icons

Soitora:
This PR adds icons to the file structure to aid the user visually in what's what.

| Before | After |
| ------ | ----- |
| image | image |

Built upon #372.

### DIFF
--- a/src/.vuepress/styles/index.styl
+++ b/src/.vuepress/styles/index.styl
@@ -134,25 +134,22 @@ body,
 	ul
 		position relative
 	li
-		font-weight bold
 		margin 0
 		padding-left 1rem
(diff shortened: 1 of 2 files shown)


arkon:
I think your other PR's changes leaked into this one?

Soitora:
Ahh yes, they were built on top of each other. They were all originally part of the same branch but there were too many changes so I would feel bad if I'd make someone review all of it...

Soitora:
I've rebased this and the other PRs now in that I dropped the leaked commits, but it's still best to do the right order of checking and merging them

ghostbear (approved):

PR #375, opened 2020-10-01
Update reader settings

ghostbear:
- Rearranged after reader setting in app
- Added "Invert Tapping"
- Added "Show content in cutout area"
- Added "Side padding"
- Added "Animate page transitions"
- Renamed "Default viewer" -> "Default reading mode"

### DIFF
--- a/src/help/guides/reader-settings.md
+++ b/src/help/guides/reader-settings.md
@@ -174,56 +174,6 @@ All examples are illustrations
 </CarouselItem>
 </Carousel>
 
-## Rotation <Badge text="Free" type="default-indicator" vertical="middle" />
-This allows you to control how the screen is going to be oriented.
-
-- **Free** lets you rotate the screen between portrait and landscape.
-- **Lock** will lock the orientation to whatever orientation you had before entering the reader.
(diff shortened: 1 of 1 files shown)


Soitora (approved):

ghostbear:
Don't merge this it seems like I missed something when doing the rebase

Soitora:
That's fine it's better to approve and let the PR author merge 👍

ghostbear:
Think I removed all duplicates. @Soitora could you give it a quick look just to double check it?

Soitora:
I can check soon when my Japanese lessons are over

Soitora:
Maybe "Page transitions" should be next to "Animate page transitions", or are they that far away in the app as well?

ghostbear:
Seems like "Page transition" has been renamed to "Animated Page transition", because I can not find it the app and they have the same description

Soitora:
Maybe cross check what options are as they should be?

ghostbear:
They all seem to be there now in order and "Page transition" was renamed to "Animated page transitions"

Soitora:
LGTM

PR #419, opened 2020-10-23
Fix assets links

FramboisePi:
This pull request fix a few non working video guide links I found when exploring Tachiyomi website. It seems their assets were renamed by #372 but their links weren't modified.


Soitora:
That's odd, I knew I changed the links during one of the commits. Thanks Lemon 👍

Soitora (approved):

PR #426, opened 2020-10-27
Update dependencies and lint for errors

Soitora:
Dependencies *and* linting in the same PR, specifically because one of the packages would fail if it wasn't linted for at the same time (I believe it was `eslint-config-vuepress` `2.2.0`->`3.0.1`).

### Changes
- Update all node packages.
- Change `element-ui` plugin tags to `PascalCase`. (Does not work with the `:::` elements though so need `kebab-case` for that still.)
- Using `slots="name"` is now outdated, so switched to `#name` which is an even newer version of `v-slot:name`.
- ESLint disabled two lines, one of which is needed (afaik) for What's New box and one of which is Vuepress native and not controlled by us.


Soitora:
Nice `Error: Cannot find module 'webpack'`, an error that did not occur for me even with a fresh `node-modules`...

Soitora:
Whyyyyyyyyyyyyyyyyyyyyyyy you do this
image

src/.vuepress/components/MaterialIcon.vue
```diff
@@ -22,7 +22,7 @@ export default {
 		},
 		icon: {
 			type: String,
-			required: true,
+			default: "",
```
ghostbear, line 25: Shouldn't this be required?

src/.vuepress/components/ForkButtons.vue
```diff
@@ -6,9 +6,11 @@
 		<button class="githubForkButton" :style="githubForkStyle" :onclick="githubForkLink">
 			{{ githubForkLabel }}
 		</button>
-		<br>
-		<span class="versionNotice">Requires <strong>Android {{androidVersion}}</strong> or higher.</span>
-	</div>	
+		<br />
+		<span class="versionNotice"
+			>Requires <strong>Android {{ androidVersion }}</strong> or higher.</span
+		>
```
ghostbear, line 12: This looks weird. I guess that the linter got confused?
```suggestion
		<span class="versionNotice">Requires <strong>Android {{ androidVersion }}</strong> or higher.</span>
```

src/.vuepress/components/MaterialIcon.vue
Soitora: Technically yeah but it conflicted with some elements that used for example :prop to get value instead of prop and that made an error to the log

src/.vuepress/components/ForkButtons.vue
Soitora: The troublesome part with the linter doing this is that any change will just revert it back unless you eslint disable the chunk it seems like

ghostbear: Could use `"htmlWhitespaceSensitivity": "ignore"` in `.prettierrc`. Which fixes this partially, it would look like this
```html
<span class="versionNotice">
  Requires
  <strong>Android {{ androidVersion }}</strong>
  or higher.
</span>
```
instead of this
```html
<span class="versionNotice"
  >Requires <strong>Android {{ androidVersion }}</strong> or higher.</span
>
```

Soitora: Sure, that looks better

ghostbear (approved):

PR #428, opened 2020-10-31
Merge two lists of common errors

### LINKED ISSUE #415: Guides: Troubleshoot - Merge two lists of common errors
The first list is at the top of https://tachiyomi.org/help/guides/troubleshooting-problems and the second list is here https://tachiyomi.org/help/guides/troubleshooting-problems/#what-are-some-common-errors . There is no reason to keep both, I suggest to remove the first one and add some errors to the second one:
- HTTP error 5xx
Most probably the site you're trying to access has problems on their side. Open the source in webview and check if the site is down.
- HTTP error 429 (Too Many Requests)
The source banned your IP address (in most cases it's temporary). We suggest to [migrate](https://tachiyomi.org/help/guides/source-migration/) part of your manga to another source.
- Unable to resolve host / Connection failed
That means that something prevents your connection with the site. Possible reasons: your internet connection is bad, the app doesn't have access to internet, your ISP have blocked the site, the site is down. Try using different internet connection (switch to Wi-Fi, mobile data or a VPN). Try to enable *More > Settings > Advanced > DNS over HTTPS*.
- java.security.cert.CertPathValidatorException / Chain validation failed
Ensure that you have right date and time on your phone. Try *More > Settings > Advanced > Clear cache/cookies*. Try using different internet connection (switch to Wi-Fi, mobile data or a VPN). Check if the site's certificate have expired - use https://www.sslshopper.com/ssl-checker.html or similar service or your browser (change URL to *__https__://example.com* or use a browser extension that forces HTTPS on sites). If the certificate have expired, then wait while the site owners will renew it. 

Also add note before second list *Note: if error-specific instruction didn't help or your error isn't in the list, go through Diagnosis*. With this note, the second list could be moved to the top and replace the first list.

Also to not create new issue, it could be added along with this one - add *More > Settings > Advanced > DNS over HTTPS* to Diagnosis

scb261:
Moved the second list in place of the first one. Removed the first list. Changed disclaimer about common errors. Closes #415


Soitora:
Accordion additions on-hold until Algolia responds to my inquiries over e-mail.

scb261:
It looks like text inside of accordions is searchable. Maybe modify the text from "It means that..." to "*Loader not implemented* means that..." and so on

Soitora:
That could work as a temporary measure?

Soitora:
Could you resolve the conflicts? Then I could merge it

scb261:
Done. Should I also reword common errors for search to work? You self-assigned, so I didn't do it

Soitora:
You can do so if you want, I only self-assigned this so that Ghost didn't need to go in and check it as well

Been a few months since I reviewed it but it should be fine overall

scb261:
I'm not sure how algolia works - it does find already merged text in accordions, but not the one in an open pr

Soitora:
It takes a while for Algolia to show it

Soitora (approved):

PR #444, opened 2020-11-11
Redirect user to GitHub instead of API when they have slow internet

ghostbear:
The issue:
https://discord.com/channels/349436576037732353/615495625231958047/775999633703960597
Could reproduce this with Chrome on Android debugging with Slow 3G 

I changed so instead of redirecting the user to the GitHub API, it redirects them to the fork latest release on GitHub.
So let's say it fails to fetch the apk download link before the user presses the download button it would redirect to https://github.com/CarlosEsco/Neko/releases/latest, instead of how it did previously it would redirect to https://api.github.com/repos/CarlosEsco/Neko/releases/latest


Soitora:
Great! 👍

Soitora (approved):

PR #450, opened 2020-11-17
Add nomedia instruction to local guide

loocool2:
As no nomedia file is automatically added like Tachiyomi does to downloads, it needs to be added manually

### DIFF
--- a/src/help/guides/reading-local-manga.md
+++ b/src/help/guides/reading-local-manga.md
@@ -11,6 +11,7 @@ Follow the steps below to create local manga.
 1. Create a folder named `local` in the `/Tachiyomi/` folder.
 	> This is located in the phone's **internal storage** or **external SD card**.
 1. Place correctly structured manga inside `/Tachiyomi/local/`.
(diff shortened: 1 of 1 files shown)


ghostbear (commented):
I would say an unordered list looks better than an ordered one. And with this optional step it could be done whenever so I don't think it needs to be ordered
With ordered list:
image
With unordered list
image

src/help/guides/reading-local-manga.md
```diff
@@ -11,6 +11,7 @@ Follow the steps below to create local manga.
 1. Create a folder named `local` in the `/Tachiyomi/` folder.
 	> This is located in the phone's **internal storage** or **external SD card**.
 1. Place correctly structured manga inside `/Tachiyomi/local/`.
+	1. Optional: If adding manga in folders, add a file named `.nomedia` to the local folder so images do not show up in the gallery
```
ghostbear, line 14: Change this to an unordered list

ghostbear (changes requested):

ghostbear (approved):

PR #455, opened 2020-11-23
Update TachiSy icon

loocool2:


arkon:
The asset seems unnecessarily large. Maybe resize it to the same size as the others?

ghostbear (approved):

PR #462, opened 2020-11-25
Add spreadsheet to the extension page

Soitora:
Added a spreadsheet table that displays @Riztard's maintained version as a tabbed item on the Extension page.

Can't say I'm completely done with it yet (@Seishirou101 is also looking into ways of enhancing i t) but it should be ready enough to be served for now.

Also stripped away `eu.kanade.tachiyomi.extension.` from the extension IDs to match the spreadsheet.

Please check my code too as it's work over several months of on and off.


Soitora:
Failed due to `vue/no-deprecated-slot-attribute` and `vue/no-deprecated-slot-scope-attribute`, will look into it.

Soitora:
For some reason, entering Extensions directly with https://deploy-preview-462--tachiyomi.netlify.app/extensions/ it will freeze the browser but entering the Extensions page from inside the website (like Sidebar) it'll load without issue.

DevTools doesn't tell me anything, anyone else able to see the issue?

ghostbear:
Install `VueGoodTable` globally in `enhanceApp.js`, it seems to fix the problem.

Could reproduce the bug on Netlify by building it with `vuepress build src` and then throwing it in an Nginx container, and then going to `localhost:8080/extensions`

And when I install `VueGoodTable` globally then I'm able to load the page directly from `localhost:8080/extensions`

```js
// ...
import 'vue-good-table/dist/vue-good-table.css'
import { VueGoodTable } from 'vue-good-table';

export default ({
	Vue, // the version of Vue being used in the VuePress app
	options, // the options for the root Vue instance
	router, // the router instance for the app
	siteData, // site metadata
}) => {
	// ...
	Vue.component("VueGoodTable", VueGoodTable);
};
```

Soitora:
Thanks @ghostbear, it seemed to work. Very odd issue

ghostbear (approved):
LGTM 👍 

There are some inconsistencies in the spreadsheet where an extension can have a lot of languages and is not `EN, SE + X more`, but that seems to be done on the spreadsheet nothing we can do about it.

PR #469, opened 2020-11-27
resized sy image

loocool2:


ghostbear (approved):

arkon:
It bothers me that it isn't a circle like the rest of them....

CrepeTF:
Good point, make all the other icons rounded rectangles

PR #470, opened 2020-11-28
Remove diacritic search

Soitora:
Increase search performance by a significant margin by removing diacritic search.

### DIFF
--- a/src/.vuepress/components/SourceSheet.vue
+++ b/src/.vuepress/components/SourceSheet.vue
@@ -4,7 +4,11 @@
 			:columns="columns"
 			:rows="items"
 			:sort-options="{ enabled: true }"
-			:search-options="{ enabled: true, placeholder: 'Search for extensions or sources' }"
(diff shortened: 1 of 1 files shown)


arkon (approved):

ghostbear (commented):
That moment when users type `Mangas Yuri` instead of `Mangás Yuri`, and complain that they can't find it

Soitora:
> That moment when users type `Mangas Yuri` instead of `Mangás Yuri`, and complain that they can't find it

Either we normalize (remove diacritics) all source names on the sheet or we just say fuck it

ghostbear:
> > That moment when users type `Mangas Yuri` instead of `Mangás Yuri`, and complain that they can't find it
> 
> Either we normalize (remove diacritics) all source names on the sheet or we just say fuck it

I can't speak for the Portuguese, but right now I think we can say fuck it. And if a lot of people complain about not finding source X, then maybe normalize it

PR #471, opened 2020-11-29
Consider multi-user mentions in 'What's new'

scb261:


Soitora (approved):

Soitora:
Great!
Works locally in a test release generated just to check this.

PR #501, opened 2021-01-12
Troubleshoot guide: common errors, CAPTCHA > Cloudflare, fix a link

scb261:
- Add Cloudlfare error (copied from error 403)
- Add one more possible reason for error 403
- Add error 404

Is it possible to see what parts of the site are referencing to https://tachiyomi.org/help/guides/troubleshooting-problems/#solving-a-captcha ? I would add Cloudfare to it, like "Solving a CAPTCHA / Cloudflare" or "Solving Cloudflare" (CAPTCHA is Cloudflare, but Cloudflare doesn't always have CAPTCHA)

### DIFF
--- a/src/help/guides/troubleshooting-problems.md
+++ b/src/help/guides/troubleshooting-problems.md
@@ -19,7 +19,7 @@ If error-specific instructions didn't help or your error isn't on the list, go t
 Often **Java.lang Exception: Challenge not found** will be resolved by force-quitting and reopening **Tachiyomi**.
 :::
 ::: el-collapse-item title="Java.lang Exception: Failed to bypass Cloudflare"
-**Java.lang Exception: Failed to bypass Cloudflare** means that the source you selected has [Cloudflare](#solving-cloudflare) protection on, refer to the [Cloudflare](#solving-cloudflare) guide to fix it.
+**Java.lang Exception: Failed to bypass Cloudflare** means that the source you selected has [Cloudflare](#solving-the-cloudflare-issue) protection on, refer to the [Cloudflare](#solving-the-cloudflare-issue) guide to fix it.
 :::
 ::: el-collapse-item title="Loader not implemented"
 **Loader not implemented** means that the respective extension for the manga is not installed. To fix, install the extension for the manga. If it still doesn't work, uninstall then reinstall the extension again.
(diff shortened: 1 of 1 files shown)


ghostbear:
> Is it possible to see what parts of the site are referencing to https://tachiyomi.org/help/guides/troubleshooting-problems/#solving-a-captcha ? I would add Cloudfare to it, like "Solving a CAPTCHA / Cloudflare" or "Solving Cloudflare" (CAPTCHA is Cloudflare, but Cloudflare doesn't always have CAPTCHA)

image
Only place that I can see refers to [troubleshooting-problems/#solving-a-captcha](https://tachiyomi.org/help/guides/troubleshooting-problems/#solving-a-captcha) are in the same file. Maybe add a subheading to `Solving a CAPTCHA` called `Couldflare CAPTCHA` and say that Cloudflare may or may not include a CAPTCHA and how to solve it

scb261:
I changed (almost) all mentions of CAPTCHA to Cloudflare outside of the *Solving Cloudflare*. Is this fine?

src/help/guides/troubleshooting-problems.md
```diff
@@ -106,11 +115,11 @@ If the site is reachable and fully functional then there may be an issue with th
 
 If the site is not reachable or having issues then all you can do is wait for the site to become functional again.
 
-## Solving a CAPTCHA
+## Solving Cloudflare
 
 **Please note that WebView is not the same as using your browser.**
 
-You must solve the **CAPTCHA** in **WebView**.
+You must solve the **Cloudflare** in **WebView**.
```
ghostbear, line 122: Are a normal CAPTCHA and a Cloudflare CAPTCHA solved in a similar way? 

I was more thinking something like this when I said add a subheading 
```md
## Solving a CAPTCHA
Lorem ipsum...

### Solving a Cloudflare CAPTCHA     <- the new subheading
Note: Cloudflare may or may not display a CAPTCHA on the extension you are using
...some more text here if needed...

**How to solve**
Explain how to solve it
```
You could then refer to `#solving-a-cloudflare-captcha`

But if they are solved in a similar way maybe just add a `note` after `Solving a CAPTCHA` saying that a Cloudflare CAPTCHA is solved in a similar way and that they should follow the same steps as a normal CAPTCHA to solve it

scb261: The app returns **Java.lang Exception: Failed to bypass Cloudflare** error in both cases and they are solved almost the same, that's why I thought it would be better to have Cloudflare (more general case) instead of CAPTCHA:
1. Open WebView
2. See Cloudflare loading
3. Cloudflare may or may not give CAPTCHA, if so - solve it
4. Wait until Clodflare disappear and the site is loaded

ghostbear (approved):
👍

PR #502, opened 2021-01-12
Update dependencies

Soitora:
Update all dependencies ~~but `eslint-config-vuepress` and the `@mr-hope/*` ones~~.


Soitora:
Don't have time to check this error:
```
4:42:49 PM: /opt/build/repo/src/.vuepress/theme/components/AlgoliaSearchBox.vue
4:42:49 PM:   42:4  error  Parsing error: Unexpected token import
```
But that's the error that made me PR everything but the `eslint-config-vuepress`

scb261:
Changing parser from default to `babel-eslint` seems to solve the problem, but I don't know if it's a good idea, maybe there is some option in `.eslintrc.js` to configure imports (I tried some but they didn't help)
```
	parserOptions: {
		...
		parser: "babel-eslint",
	},
```
> You only need to use babel-eslint if you are using types (Flow) or experimental features not supported in ESLint itself yet. Otherwise try the default parser (you don't have to use it just because you are using Babel).

Soitora:
> Changing parser from default to `babel-eslint` seems to solve the problem

Yeah I saw that fix too, althought I was running it with some experimental flags at that time so it gave some other errors, running it "vanilla" worked fine, so applying that.

ghostbear (approved):

PR #506, opened 2021-01-15
Update FAQ

scb261:
- There is no **Refresh library metadata** option anymore, but instead there is **Refresh library manga covers** now.
- The app now allows parallel downloads, but only from up to 5 different sources.
- New search method in MAL (by id).
- The Tachiyomi folder isn't automatically created on install and requires some actions to be automatically created, but the actions in faq aren't good examples: download location and automatic backup location can be previously changed by the user and manual backup suggests last selected folder. I only left downloading (and added extra steps to ensure that the folder will be created in the right place), since auto-backup requires waiting and there is no default folder to revert to.
- Guide how to fix missing covers for local manga (the **Refresh library metadata** option doesn't help in this case).
- Fixed few small mistakes.

### DIFF
--- a/src/help/faq/README.md
+++ b/src/help/faq/README.md
@@ -93,10 +93,10 @@ Set download location to default by going to <Navigation item="more"/> → <Navi
 Sometimes some covers for local manga aren't displayed. Follow these steps to fix it:
 
 ::: guide
-1. Make sure you've created right folder structure. To do it, open the manga with missing cover and check if you can read chapters in the app. If not, follow [this](/help/guides/reading-local-manga) guide first.
-1. Remember which chapters you've read and remove manga from library.
+1. Make sure you've created the right folder structure. To check it, open the manga with the missing cover and check if you can read chapters in the app. If not, follow [this](/help/guides/reading-local-manga) guide first.
+1. Remember, or note down, which chapters you've read and remove manga from the library.
 1. Go to <Navigation item="more"/> → <Navigation item="settings"/> → <Navigation item="settings_advanced"/> and tap **Clear database**. This only affects manga that aren't in your library.
-1. Go to <Navigation item="browse"/> → **Local source** and find the manga, the cover should be fixed now. Add manga back to library, mark read chapters and add tracking if needed.
+1. Go to <Navigation item="browse"/> → **Local source** and find the manga, the cover should be fixed now. Add manga back to the library, mark read chapters and add tracking if needed.
 :::
 
 ::: note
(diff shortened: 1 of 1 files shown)


src/help/faq/README.md
```diff
@@ -53,8 +53,11 @@ If the website/scanlator group doesn't have an online manga reader, you can down
 Aside from network issues, it could be because the images are too big or that the decoder doesn't support that image type.
 
 ### Downloads
-#### What happened to the simultaneous downloads feature?
-It was removed to prevent IP bans caused by too many requests. We know that speed is being sacrificed but it's better than not being able to access a source at all.
+#### How to download several chapters or manga simultaneously?
+
+If you're downloading several manga, you can use different sources. In this case the app will start downloading simultaneously from up to 5 different sources.
+
+The app doesn't allow parallel downloads from one source to prevent IP bans caused by too many requests. We know that speed is being sacrificed but it's better than not being able to access a source at all.
```
ghostbear, line 60: Maybe swap this around

```suggestion
#### How to download several chapters or manga simultaneously?
The app doesn't allow parallel downloads from one source to prevent IP bans caused by too many requests. We know that speed is being sacrificed but it's better than not being able to access a source at all.

However, if you're downloading several manga, and they use different sources. In this case, the app will start downloading simultaneously from up to 5 different sources.
```

src/help/faq/README.md
```diff
@@ -80,13 +83,31 @@ Related GitHub issue: [#65](https://github.com/tachiyomiorg/tachiyomi/issues/65)
...
+:::
+
+#### How to fix empty cover for a local manga?
+Sometimes some covers for local manga aren't displayed. Follow these steps to fix it:
+
+::: guide
+1. Make sure you've created right folder structure. To do it, open the manga with missing cover and check if you can read chapters in the app. If not, follow [this](/help/guides/reading-local-manga) guide first.
+1. Remember which chapters you've read and remove manga from library.
+1. Go to <Navigation item="more"/> → <Navigation item="settings"/> → <Navigation item="settings_advanced"/> and tap **Clear database**. This only affects manga that aren't in your library.
+1. Go to <Navigation item="browse"/> → **Local source** and find the manga, the cover should be fixed now. Add manga back to library, mark read chapters and add tracking if needed.
```
ghostbear, line 99:
```suggestion
1. Make sure you've created the right folder structure. To check it, open the manga with the missing cover and check if you can read chapters in the app. If not, follow [this](/help/guides/reading-local-manga) guide first.
1. Remember, or note down, which chapters you've read and remove manga from the library.
1. Go to <Navigation item="more"/> → <Navigation item="settings"/> → <Navigation item="settings_advanced"/> and tap **Clear database**. This only affects manga that aren't in your library.
1. Go to <Navigation item="browse"/> → **Local source** and find the manga, the cover should be fixed now. Add manga back to the library, mark read chapters and add tracking if needed.
```

ghostbear (approved):
👍

PR #524, opened 2021-01-29
Troubleshoot: WebView rewrite, minor changes

scb261:
- Add updating WebView to diagnosis 
- Punctuation
- Bold **Update** button and **Extensions** tab
- Rewrite WebView part at the end

### DIFF
--- a/src/help/guides/troubleshooting-problems.md
+++ b/src/help/guides/troubleshooting-problems.md
@@ -141,9 +141,9 @@ To update **WebView**, you need to find what **WebView implementation** is used 
 * Android 6.0.1 and lower - [Android System WebView](https://play.google.com/store/apps/details?id=com.google.android.webview).
 
 ::: c-tip
-Users of Android 7.0 and higher can also see and change **WebView implementaion** in [Developer Options](https://developer.android.com/studio/debug/dev-options).
+Users of Android 7.0 and higher can also see and change **WebView implementation** in [Developer Options](https://developer.android.com/studio/debug/dev-options).
 :::
 
 ::: c-warning
-Non-standard implementations like Firefox can lead to incorrect work or crash of Tachiyomi. We recommend to set **WebView implementaion** to [Android System WebView](https://play.google.com/store/apps/details?id=com.google.android.webview) or [Google Chrome](https://play.google.com/store/apps/details?id=com.android.chrome).
-
+Non-standard implementations like Firefox can lead to incorrect work or crash of Tachiyomi. We recommend to set **WebView implementation** to [Android System WebView](https://play.google.com/store/apps/details?id=com.google.android.webview) or [Google Chrome](https://play.google.com/store/apps/details?id=com.android.chrome).
+:::
--- a/src/help/guides/troubleshooting-problems.md
+++ b/src/help/guides/troubleshooting-problems.md
@@ -145,5 +145,5 @@ Users of Android 7.0 and higher can also see and change **WebView implementation
 :::
 
 ::: c-warning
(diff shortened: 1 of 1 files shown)


Seishirou101:
I believe depending on your android version, what is used as webview by the device is dependent that. Apparently android webview is the standard for 7.0 and below, chrome being the default/forced for [7.0-10](https://www.androidpolice.com/2016/07/20/google-explains-chrome-will-become-webview-android-7-0/), and after 10, a new one called [Trichrome](https://www.xda-developers.com/google-chrome-no-longer-webview-provider-android-10/). So maybe mentioning the others may be needed, I'm unsure even how you would update the new one, whereas for 7-10 you'd just update chrome.

scb261:
So should it just be like
- A5-6 (link to Android System WebView)
- A7-9 (link to chrome)
- A10 and above (link to Trichrome)

Or should there be a guide to check what webview is used (since it can be changed in settings and I'm not sure if all manufacturers are following the list above by default)

Seishirou101:
It's generally forced, at least from what I've seen. You'll have to root to change it or remove the webview to change it (in my case I had to uninstall chrome) but I may be wrong. 
Screenshot_20210129-123905136

scb261:
I'm thinking about rewriting https://tachiyomi.org/help/guides/troubleshooting-problems/#webview-not-working part.
- Make it at the same level as Cloudflare (right now WebView is part of Cloudflare)
- Change title to just *WebView*
- Make two sub-titles *Update WebView* and *Check what WebView implementation is used*. In the first one list of the three links with android versions and note that it can be different in user's case. In the second one tell about developer options, tell that other implementations may crash or work incorrectly with Tachiyomi
- Link *Update WebView* in the Diagnosis

scb261:
Ready for review, I think

src/help/guides/troubleshooting-problems.md
```diff
@@ -132,13 +133,17 @@ You must solve the **Cloudflare** issue in **WebView**.
 1. Once done, press the *Back Arrow* at the top left to return.
 :::
 
-### WebView not working?
-If **WebView** doesn't work for you, please update the [Android System WebView](https://play.google.com/store/apps/details?id=com.google.android.webview) app.
+## Update WebView
 
-When you've updated go to [Developer Options](https://developer.android.com/studio/debug/dev-options) in your phone's settings and set **WebView Implementation** to `Android System WebView`.
+To update **WebView**, you need to find what **WebView implementation** is used on your device. In general, default implementation depends on Android version as follows:
```
ghostbear, line 138:
```suggestion
To update **WebView**, you need to find what **WebView implementation** is used on your device. In general, default implementation depends on the Android version as follows:
```

src/help/guides/troubleshooting-problems.md
```diff
@@ -132,13 +133,17 @@ You must solve the **Cloudflare** issue in **WebView**.
...
+* Android 6.0.1 and lower - [Android System WebView](https://play.google.com/store/apps/details?id=com.google.android.webview).
 
 ::: c-tip
-If `Android System WebView` doesn't work, try setting `Google Chrome` as the default **WebView Implementation** instead.
+Users of Android 7.0 and higher can also see and change **WebView implementation** in [Developer Options](https://developer.android.com/studio/debug/dev-options).
 :::
 
-
+::: c-warning
+Non-standard implementations like Firefox can lead to incorrect work or crash of Tachiyomi. We recommend to set **WebView implementation** to [Android System WebView](https://play.google.com/store/apps/details?id=com.google.android.webview) or [Google Chrome](https://play.google.com/store/apps/details?id=com.android.chrome).
```
ghostbear, line 148: A bit hard to read, maybe something like this
```suggestion
Non-standard implementations like Firefox can lead to Tachiyomi not working correctly or in the worst-case crash. We recommend to set **WebView implementation** to [Android System WebView](https://play.google.com/store/apps/details?id=com.google.android.webview) or [Google Chrome](https://play.google.com/store/apps/details?id=com.android.chrome).
```

ghostbear (approved):
LGTM 👍

PR #533, opened 2021-02-07
Update unique features from forks

arkon:
Now that various features have made their way across to the main app and other forks.


src/forks/README.md
```diff
@@ -41,7 +41,7 @@ This hentai focused fork is based on **[TachiyomiEH](/forks/TachiyomiEH)**, it h
 
 Some features include:
 * Built in sources of many popular hentai sources with additional features.
-* Manga recommendations from My Anime List and AniDB
+* Manga recommendations from MyAnimeList and Anilist
```
ghostbear, line 44: Nitpick 
```suggestion
* Manga recommendations from MyAnimeList and AniList
```
I also assume this is true due to SY having the same phrase

ghostbear (approved):
LGTM 👍

PR #539, opened 2021-02-09
add download-migration warning

### LINKED ISSUE #535: Guides: Clarify the state of downloaded manga when migrating to another source
Migrating to a different source does not migrate downloads, and unless user deletes the files are orphaned. Thus user should either delete existing downloads from the current source first then redownload from migrated source, or move the downloads to a local manga to read offline.

This isn't particularly intuitive for users nor its mentioned in the app (the orphan part especially)

Seishirou101:
closes #535

### DIFF
--- a/src/help/guides/source-migration.md
+++ b/src/help/guides/source-migration.md
@@ -13,9 +13,9 @@ Always make sure to have a [backup](/help/guides/creating-backups/) in case anyt
 :::
 
 ::: c-warning
-Downloaded chapters do not transfer with migrations.
+Downloaded chapter(s) do not transfer with migrations.
 
-Migrations with downloaded chapters may leave the download behind. You will need to remove these manually with a file manager.
+Migrations with downloaded chapter(s) may leave the download behind. You will need to remove these manually with a file manager.
 :::
 
 ## Migrating on Stable/Preview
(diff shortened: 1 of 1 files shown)


ghostbear (approved):
LGTM :+1: 

But it is a bit hard to read

Seishirou101:
> 
> 
> LGTM 👍
> 
> But it is a bit hard to read

hmmm, maybe something like this is better then? 
change

ghostbear:
> > LGTM +1
> > But it is a bit hard to read
> 
> hmmm, maybe something like this is better then?
> change

```
Downloaded chapter do not transfer with migrations.

Migrations with downloaded chapter may leave the download behind. You will need to remove these manually with a file manager.
```
Maybe?

Seishirou101:
> 
> 
> > > LGTM +1
> > > But it is a bit hard to read
> > 
> > 
> > hmmm, maybe something like this is better then?
> > change
> 
> ```
> Downloaded chapter do not transfer with migrations.
> 
> Migrations with downloaded chapter may leave the download behind. You will need to remove these manually with a file manager.
> ```
> 
> Maybe?

Yeah, I think that sounds good. I did downloaded chapter(s) without the parentheses though. Or do you think I should leave them singular?

ghostbear:
There will probably be one or more downloaded chapters so doing parentheses is probably better

Seishirou101:
sounds good, it has been added

PR #561, opened 2021-03-03
Shows forks on sidebar

specterflare:


specterflare:
Plz review

ghostbear (approved):
I'm fine with this. I will let @Soitora have the final say in this though

Soitora:
Also fine since we're already doing this on V2

PR #570, opened 2021-03-11
Sidebar changes and renaming some Guide items

Soitora:
This brings v1 closer to v2 in terms of how the sidebar is designed. Also since some of these guide items are changed in v2 we might as well just do the rename on them now for the redirect.


ghostbear (approved):
LGTM 👍

PR #571, opened 2021-03-12
Fix extensions page

Soitora:
Backports most of the V2 code (but not styling) for Extensions to V1.

Also fixes a functionality that should have always worked but never did (extension selection).

Search is not in V2 yet but probably won't be backported when it is added

### DIFF
--- a/src/.vuepress/components/ExtensionList.vue
+++ b/src/.vuepress/components/ExtensionList.vue
@@ -1,26 +1,41 @@
 <template>
-	<div>
+	<div class="extension-list">
 		<div v-for="extensionGroup in extensions" :key="extensionGroup[0].lang">
-			<h3>{{ langName(extensionGroup[0].lang) }}</h3>
+			<h3>
+				{{ langName(extensionGroup[0].lang) }}
+				<span class="extensions-total">
+					Total:
+					<span class="extensions-total-sum">
+						{{ extensions.reduce((sum, item) => sum + item.length, 0) }}
+					</span>
+				</span>
+			</h3>
 			<div
 				v-for="extension in extensionGroup"
-				:id="extension.name.split(': ')[1]"
+				:id="extension.pkg.replace('eu.kanade.tachiyomi.extension.', '')"
 				:key="extension.apk"
 				class="anchor"
 			>
 				<div class="extension">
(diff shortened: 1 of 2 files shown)


src/.vuepress/components/ExtensionList.vue
```diff
@@ -47,7 +61,30 @@ export default {
 	async beforeMount() {
 		const { data } = await axios.get(GITHUB_EXTENSION_JSON);
 		const values = Object.values(groupBy(data, "lang"));
-		this.$data.extensions = sortBy(values, [(g) => this.langName(g[0].lang)]);
+		values.sort((a, b) => {
+			const langA = this.langName(a[0].lang).split(" ")[0];
+			const langB = this.langName(b[0].lang).split(" ")[0];
```
ghostbear, line 66: Think this would be cleaner 
```suggestion
			const langA = ISO6391.getName(a[0].lang);
			const langB = ISO6391.getName(b[0].lang);
```

Soitora: Added it but what difference was this to the existing `langName` function?

ghostbear: The thing with the existing `langName` function is that it returns the language name plus the language name in its native language with parentheses e.g. `Russian (Русский)`

The difference with the `simpleLangName` function is that it will just return the language name e.g. `Russian`

So instead of splitting the string, `Russian (Русский)`, and then getting the language name by picking the first element in the array, `Russian`, we just get the language name from `ISO6391.getName(code)` with the language code `ru` which then just returns `Russian`

src/.vuepress/components/ExtensionList.vue
```diff
@@ -1,26 +1,41 @@
 <template>
-	<div>
+	<div class="extension-list">
 		<div v-for="extensionGroup in extensions" :key="extensionGroup[0].lang">
-			<h3>{{ langName(extensionGroup[0].lang) }}</h3>
+			<h3>
+				{{ langName(extensionGroup[0].lang).replace("English (English)", "English") }}
```
ghostbear, line 5: You could probably just check if the language code is "en" then just change the function that is being called
```suggestion
				{{ extensionGroup[0].lang === "en" ? simpleLangName(extensionGroup[0].lang) : langName(extensionGroup[0].lang) }}
```

ghostbear (approved):

PR #573, opened 2021-03-12
Fix redirects

Soitora:
Maybe fix redirects?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
 	"requires": true,
 	"packages": {
 		"": {
-			"name": "tachiyomi-website",
 			"version": "1.0.0",
 			"license": "MPL-2.0",
 			"dependencies": {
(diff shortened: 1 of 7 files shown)


Soitora:
Hmm still redirects but gives `404` until page is refreshed, in local enviroment this works flawlessly. Maybe use Cloudflare for 301 for Permanent Redirects? Or anyone knows another solution to 301 redirecting?

arkon:
Can we not just update whatever old references there are and not have redirects?

Soitora:
> Can we not just update whatever old references there are and not have redirects?

We can, but Google will whine on us but I guess that's fine

Soitora:
> Can we not just update whatever old references there are and not have redirects?

We can, but Google will whine on us but I guess that's fine

Soitora:
Seems that doing a simple redirect then using dehydrate will fix the 404s

arkon (approved):

arkon:
> Rebuilding package-lock didn't solve conflict, what's the best way to fix?

Did you rebase on top of master first?

arkon (approved):

PR #574, opened 2021-03-13
Change top margin and padding for h2

Soitora:
Don't know where along the lines this was changed but it failed to take in account for the navbar when going to linked titles. This fixes that.


ghostbear (approved):
LGTM 👍

PR #575, opened 2021-03-13
Update to NPM 7 in build script

ghostbear:
This:
- updates NPM to v7
- adds a cache for dependencies

We had some issues with NPM 6 on the rewrite project and NPM 7 solved the issues we experienced. So to not have any more conflicts with NPM we thought we might as well update to NPM 7 here too

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -12,15 +12,27 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
 
-      - name: Install Node v12
-        uses: actions/setup-node@v1
+      - name: Setup Node 12
+        uses: actions/setup-node@v2
         with:
           node-version: 12
+          
+      - name: Setup NPM 7
+        run: npm install -g npm@7
+        
+      - name: Check cache
+        uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


Soitora (approved):
I know it worked for ours so I can approve, but maybe @arkon has something else to say

arkon:
Maybe we should upgrade Node too?

Soitora:
There are two dependencies that yell the following:
```js
npm WARN deprecated fsevents@1.2.13: fsevents 1 will break on node v14+ and could be using insecure binaries. Upgrade to fsevents 2.
npm WARN deprecated chokidar@2.1.8: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
```
Which is why we kept it at NodeJS 12 for now, and NodeJS 13 seems like a weird upgrade to do

arkon:
Isn't LTS 14 or something now? It was just a thought since a newer version might have npm7 bundled anyway.

This is fine as-is though.

ghostbear:
Node 14, which is LTS, still includes NPM 6

arkon (approved):

PR #587, opened 2021-03-16
Remove @mr-hope/last-updated dependency in favour for the built in one

### LINKED ISSUE #563: Pages are always "Last updated: a minute ago"
No matter what, every page on the site says "Last updated: a minute ago". I'm not sure when it broke, but I do have distinct memories of it not being broken.

ghostbear:
Will this fix #563 maybe it works locally with both `serve` and `build`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,10 +5,10 @@
 	"requires": true,
 	"packages": {
 		"": {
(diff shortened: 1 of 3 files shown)


Soitora:
We'll see

Soitora (approved):

PR #627, opened 2021-05-06
update j2k banner

### LINKED ISSUE #625: update j2k page with new app screenshots
as of v1.0.16, multiple changes have been made including the addition of the new toolbar

OncePunchedMan:
Closes #625


arkon:
We probably want to avoid showing any specific sources.

OncePunchedMan:
> We probably want to avoid showing any specific sources.

i chose mangaplus as source since it's official. also the sources on extension tab aren't the installed ones

Soitora:
> > We probably want to avoid showing any specific sources.
> 
> i chose mangaplus as source since it's official.

That could be an issue (especially since on F-droid I *think* they mentioned seeing the Webtoons.com source), which is why the new vanilla banner looks like this:
App banner

OncePunchedMan:
so no extension tab anymore and local source in manga details page?

Soitora:
Perhaps, that's the safest bet

arkon:
Whoops I forgot to respond, but yes to what Soi said.

We should update the other images too, including the one on the main page.

ghostbear:
There are some videos where we show how users install/use extensions. Should we also update the video to use something that is maybe self-hosted like Komga?

OncePunchedMan:
correct, if we decide to completely remove sources then we should update all banners

arkon (approved):

Soitora (approved):
Three approvals, why not

PR #629, opened 2021-05-06
add note about rar5

### LINKED ISSUE #614: add a Note about RAR5 in Guide for Local manga
## The Problem
while local can support most RAR/CBR formats, it can't handle the newer RAR5 format, since the app doesn't support it yet. 
some users have had issues with the same and have reported at #support

## Requirement
add a Note telling users to not use RAR5. If they do have such files, they'll have to extract them and re-compress it to one of the supported formats

OncePunchedMan:
Closes #614

### DIFF
--- a/src/help/guides/local-manga.md
+++ b/src/help/guides/local-manga.md
@@ -27,9 +27,9 @@ If you add more chapters then you'll have to manually refresh the chapter list (
 Supported chapter formats are folder with pictures inside (such as `.jpg`, `.png`, etc), `ZIP`/`CBZ`, `RAR`/`CBR` and `EPUB`. But expect better performance with directories and `ZIP`/`CBZ`.
 
 ::: c-warning
-While Tachiyomi can support **RAR** chapter formats please note, that the newer **RAR5** format isn't supported yet.
+While Tachiyomi does support chapters compressed as **RAR** or **CBR**, note that **RAR** or **CBR** files using the **RAR5** format are not supported yet.
 
-If you already have **RAR5** files you need to extract and re-compress them into one of the supported formats. 
+If **CBR** or **RAR** files are not working, you may need to extract and re-compress them into one of the supported formats. 
 :::
 
 ## Folder Structure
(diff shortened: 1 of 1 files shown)


src/help/guides/local-manga.md
```diff
@@ -26,6 +26,12 @@ Supported chapter formats are folder with pictures inside (such as `.jpg`, `.png
 Remember to give the app storage permissions on **Android 6** and newer.
 :::
 
+::: c-tip
+While Tachiyomi can support **RAR** chapter formats please note, that the newer **RAR5** format isn't supported yet.
+
+If you already have **RAR5** files you need to extract and re-compress them into one of the supported formats. 
+:::
+
```
ghostbear, line 34: Maybe we should move `Remember to give the app storage permissions on **Android 6** and newer.` to be below the `Local manga` heading then change this from `c-tip` to `c-warning`

loocool2:
```suggestion
::: c-tip
While Tachiyomi does support chapters compressed as **RAR** or **CBR**, note that **RAR** or **CBR** files using the **RAR5** format are not supported yet.

If **CBR** or **RAR** files are not working, you may need to extract and re-compress them into one of the supported formats. 
:::

```

loocool2 (dismissed):
Rephrased to be clearer and include CBR files in the warning

src/help/guides/local-manga.md
OncePunchedMan: i thought CBR format is working tho:

_Supported chapter formats are folder with pictures inside (such as .jpg, .png, etc), ZIP/CBZ, RAR/CBR and EPUB. But expect better performance with directories and ZIP/CBZ._

src/help/guides/local-manga.md
```diff
@@ -6,6 +6,8 @@ lang: en-US
 
 # Local manga
 
+
```
Soitora, line 9: This and line 10, why?

Soitora (commented):
Thonk

PR #652, opened 2021-05-12
refactored sort filter

yegorgunko:
since extensions are already sorted, just reverse them if DESC selected


ghostbear (approved):
LGTM 👍 

Also, I would advise you to use branches that way you can sync your master branch with upstream without needed to force push or recreate your fork

yegorgunko:
Yeah, will do for sure...
I just thought that since it's a small and one commit, it would be good on `master`, but it is where the things started to go wrong 😅

PR #684, opened 2021-06-10
Increased extensions list performance

yegorgunko:
Performance has been improved thanks to [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) and a package [vue-virtual-scroller](https://github.com/Akryum/vue-virtual-scroller) which is a wrapper for observer.

### DIFF
--- a/src/.vuepress/components/Navigation.vue
+++ b/src/.vuepress/components/Navigation.vue
@@ -44,6 +44,7 @@ export default {
 				filter: { text: "Filter", icon: "filter_list" },
 				edit: { text: "EDIT", icon:"edit" },
 				delete: { text: "DELETE", icon:"delete" },
+				language: { text: "Filter", icon:"translate" },
 				/* More */
 				downloaded_only: { text: "Downloaded only", icon: "cloud_off" },
 				download_queue: { text: "Download queue", icon: "get_app" },
--- a/src/help/faq/README.md
+++ b/src/help/faq/README.md
@@ -9,6 +9,27 @@ sidebarDepth: 2
 
 ## Application
 
+### General
+
+#### Will there be an iOS version?
+There is no iOS version, and there are no plans for one. Since iOS and Android apps do not share code, it is hard to port.
+
+#### Will Tachiyomi ever be on the Google Play Store?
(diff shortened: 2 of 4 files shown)


yegorgunko:
@ghostbear 'virtual scroll' should remove elements that aren't currently visible freeing up the DOM tree and thus boosting lists performance.
The hard thing in our particular case is groups. Since they have own loop. This package doesn't work well sometimes because of that. But as far as I could see, it doesn't keep all of the elements in the DOM which is it's purpose.

I will look into ways of implementing virtual scroll without external packages and for it to be able to work with our use case: groups.

yegorgunko:
Well, I investigated a lot about virtual scrolling technique and I can say it is quite complicated.
For now I'm thinking of using a method that relies on absolute positioning / top padding (thanks [Vuetify](https://github.com/vuetifyjs/vuetify) for initial idea). To summarize what can be done in order to achieve virtual scroll:

1. Creating a container which will be overflowed and which will have `relative` positioning
2. Creating a list container which has a height of `totalItems * itemHeight`
3. Creating a list items which will have `absolute` positioning and will be generated by Vue having dynamic `top` style. Their top offset will be calculated based on the `scrollTop` value of the list.

By using such an approach, the DOM tree will have only the elements that are in the viewport while preserving the original list length and making it possible to scroll.
So, in general: for the user it will look just like an ordinary list with a scrollbar, but in the DOM there will be only a few elements rendered.

I'm finding it quite hard to implement for now, but I will look into it more.

yegorgunko:
@ghostbear I've created a sandbox with my own virtual scroll implementation
It shows exactly what virtual scroll is about. It features a list with 100000 items.

https://codesandbox.io/s/vue-virtual-scroll-lemle

ghostbear:
Great thanks helps a lot with understanding what it does :+1: 

Just so you know the reason why I asked if it can be done without a package is because we are currently rewriting the site and it doesn't use the same framework. So it would just add more work if another package was added and it would need to be rewritten to work with it

yegorgunko:
@ghostbear is it gonna still be Vue? Maybe Vuepress v2?

ghostbear:
It gonna still be Vue but with Gridsome instead of Vuepress

yegorgunko:
Well, in the end, there is an issue with calculating dynamic height of the list.
It is caused by async data retrieval. Even `$nextTick` doesn't work, so for now I don't know how to implement the scroll in current situation.

Maybe on a new framework base there will be a clean way to do it.

PR #686, opened 2021-06-12
Replace ” with "

ghostbear:
Found some broken tags on the website. The reason why they are broken is that they use the wrong quotation mark


Soitora:
Whip @loocool2 🎊

Soitora (approved):

loocool2:
Bruh I included this in my updated PR

PR #689, opened 2021-06-13
Use browser level lazy loading in Extension List

ghostbear:
- Adds lazy loading to images
- Split up ExtensionList into smaller components

Reduces initial load size from ~5MB down to ~1MB

When cleaning up I fixed some styles, removed unnecessary tags, and used computed properties instead of doing them in the templates

TODO:
- [x] ~~Activate polyfill library~~
- [x] Remove old list compontent


arkon:
Couldn't you just use https://web.dev/browser-level-image-lazy-loading/ ?

ghostbear:
I could but IntersectionObserver has greater browser support than browser level image lazy loading

https://caniuse.com/loading-lazy-attr

https://caniuse.com/intersectionobserver

arkon:
It's only really missing Safari and IE though, which don't really matter.

Soitora (approved):

arkon (approved):

PR #701, opened 2021-06-27
Download universal APK instead of first one for preview releases 

ghostbear:


src/.vuepress/components/DownloadButtons.vue
```diff
@@ -42,7 +42,7 @@ export default {
 		}
 		try {
 			const { data } = await this.$store.dispatch("getPreviewReleaseData");
-			const apkAsset = data.assets.find((a) => a.name.includes(".apk"));
+			const apkAsset = data.assets.find((a) => /^tachiyomi-r\d{4}.apk/.test(a.name));
```
arkon, line 45: This will break the day we hit 10000 commits (although that's pretty far away). It should be safe to just rely on the `tachiyomi-r` prefix.

Soitora: > This will break the day we hit 10000 commits (although that's pretty far away). It should be safe to just rely on the `tachiyomi-r` prefix.

That should be fine.
```suggestion
			const apkAsset = data.assets.find((a) => /^tachiyomi-r\d{4,5}.apk/.test(a.name));
```

src/.vuepress/components/DownloadButtons.vue
```diff
@@ -42,7 +42,7 @@ export default {
 		}
 		try {
 			const { data } = await this.$store.dispatch("getPreviewReleaseData");
-			const apkAsset = data.assets.find((a) => a.name.includes(".apk"));
+			const apkAsset = data.assets.find((a) => /^tachiyomi-r\d{4,}.apk/.test(a.name));
```
ghostbear, line 45: This will now make sure it's 4 or more digits that way it doesn't break when we hit 10 000, 100 000, or 1 000 000

arkon (approved):

PR #716, opened 2021-08-03
Fix pt-br returning ()

ghostbear:


arkon (commented):
Maybe we should try to just handle any country-specific locale?

ghostbear:
> Maybe we should try to just handle any country-specific locale?

It now handles country-specific locale. And the ISO-639-1 library is replaced with the browser built-in [Intl.DisplayNames](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames/DisplayNames) which seems to handle far more country codes

arkon (approved):

PR #726, opened 2021-08-11
Update home page images

ghostbear:


Soitora:
Should the extensions description be rewritten to better hint at the extensions system so people don't look at the image and see only 4 unknown sources?

Maybe replace Manga preview with manga info page or something, or the trackers page for a series?

ghostbear:
> Maybe replace Manga preview with manga info page or something, or the trackers page for a series?

Can replace it with the track sheet

Soitora:
Now the texts doesn't match the images, I'm wondering though if there's a better reader image to represent it than some 1s and 0s heart

ghostbear:
How about swapping extension and tracking text, would that be better?

arkon:
Is it just me, or is the positioning of the 2 tracking image slightly off?

ghostbear:
> Is it just me, or is the positioning of the 2 tracking image slightly off?

Fixed

arkon (approved):

Soitora (approved):

Issue #730, opened 2021-08-25
Assets that may need updates


Seishirou101:
**Pictures that need updates**

- [x] guides_nav-edge-light.png #732 
- [x] guides_nav-kindlish-light.png #732 
- [x] guides_nav-l-light.png #732 
- [x] guides_nav-right-and-left-light.png #732 
- [ ] guides_reader-background-black.png
- [ ] guides_reader-background-gray.png
- [ ] guides_reader-background-white.png
- [ ] guides_reader-crop-border-off.png
- [ ] guides_reader-crop-border-on.png
- [ ] guides_reader-fit-height.png
- [ ] guides_reader-fit-screen.png
- [ ] guides_reader-fit-width.png
- [ ] guides_reader-original-size.png
- [ ] guides_reader-smart-fit.png
- [ ] guides_reader-stretch.png
- [ ] guides_reader-zoomstartpos-center.png
- [ ] guides_reader-zoomstartpos-left.png
- [ ] guides_reader-zoomstartpos-right.png

**Pictures that need to change format to adhere to contributing guidelines and update**

- [ ] guides_webview-open.jpg
- [ ] guides_local-manga.jpg

**Videos that need to change format to adhere to contributing guidelines and update**

- [x] guides_migrate-az.mp4 #734 
- [x] guides_migrate-j2k-single.mp4 #734 
- [x] guides_migrate-j2k.mp4 #734 
- [x] guides_migrate-preview-manga.mp4 #734 
- [x] guides_migrate-preview-source.mp4 #734 
- [x] guides_migrate-sy-library.mp4 (no longer needs this, removed) #734 
- [x] guides_migrate-sy-source.mp4 #735
- [x] guides_add-to-tracker.mp4  #735

**Videos that need to update**

- [x] guides_category-add-to.webm #735
- [x] guides_category-remove-from.webm #735
- [x] guides_library-add.webm #735

**Misc assets that may need updating**

- [ ] home_library-dark.png (Use Merryweather comics for prettier images?)
- [ ] home_library-light.png
- [ ] home_tracking-dark.png
- [ ] home_tracking-light.png
- [x] guides_login-to-tracker.mp4 #735
- [x] faq_mangadex-no-results.mp4 (removed) #735

S-0-a:
Enable online saving of library through logging in our Gmail account.

PR #735, opened 2021-08-28
update last videos assets and remove not needed video asset

Seishirou101:
finishes up videos listed in #730
removed faq_mangadex-no-results.mp4 as I don't think its needed/no longer referenced.


ghostbear (approved):
LTGM 👍🏽

PR #739, opened 2021-09-06
add id searching to extension list

Seishirou101:
adds ID searching and baseurl. Not sure if the self hosted one's baseurl should be removed as that might cause confusion to *some* people but who knows.

Also not sure if we should do hyperlinks? but that can easily be changed if its not wanted.

### DIFF
--- a/src/.vuepress/components/Extensions.vue
+++ b/src/.vuepress/components/Extensions.vue
@@ -19,7 +19,6 @@ export default {
 		return {
 			extensions: [],
 			filters: {
-				searchID: "",
 				search: "",
 				lang: [],
 				nsfw: "Don't care",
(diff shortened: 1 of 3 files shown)


src/.vuepress/components/Extensions.vue
```diff
@@ -42,6 +43,11 @@ export default {
 						ext.name.toLowerCase().includes(filters.search.toLowerCase())
 					);
 				}
+				if (filters.searchID) {
+					filteredGroup = filteredGroup.filter((ext) =>
+						ext.sources.some((idnum) => idnum.id.includes(filters.searchID))
+					);
+				}
```
ghostbear, line 50: Pretty sure the search field could be reused for this

```js
if (filters.search) {
  filteredGroup = filteredGroup.filter((ext) =>
    ext.name.toLowerCase().includes(filters.search.toLowerCase()) ||
    ext.sources.some((idnum) => idnum.id.includes(filters.searchID))
  );
}
```

Seishirou101: I initially did this but for some reason thought they may need to filter out names on top of ID... which now makes no sense. will change it o7

ghostbear (approved):
I'm fine with the filter changes. 

But I'm not sure about adding `baseUrl`.

arkon:
I'd personally prefer not readily exposing the links like that.

Seishirou101:
> 
> 
> I'd personally prefer not readily exposing the links like that.

okay, removed

arkon (approved):

PR #827, opened 2022-01-16
Added more statuses to template

zayin02:


arkon:
This shouldn't be merged until after the next stable app release.

PR #828, opened 2022-01-16
Removed the width atribute

AnujSsStw:
Because of the width attribute, Tachiyomi Text moves dow
Screenshot (78)
After Removing the width 
Screenshot (79)
n


arkon (approved):

arkon:
Thanks!

PR #912, opened 2022-05-30
Add MangaUpdates to list of trackers + reorder trackers by popularity

specterflare:
Popularity determined by site rank on https://tranco-list.eu/

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -14,7 +14,7 @@ features:
     details: Online and offline reading from over a thousand sources
     image: library
   - title: Tracking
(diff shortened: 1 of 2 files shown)


arkon:
To be merged after the next stable release.

MaouArata:
Add MangaBTT.com because it has many manga / manhwa that is faster to put new chapter than others extension . For example, Reality Quest is only of chapter 44 as every extension in tachiyomi where in Mangabtt , it has up to 47 chapter . So plz add mangbtt.com

PR #952, opened 2022-10-05
Update Cloudflare troubleshooting guide

### LINKED ISSUE #940: Cloudflare detection bypass working suggestion
### Describe why this should be added

I was able to discover a way to force bypass cloudflare errors via suggesting this feature "Clear Webview Data" inside advanced settings of Tachiyomi. https://github.com/tachiyomiorg/tachiyomi/commit/6e95fde4ece64e5959c04bb4b7fb69299ef37ca5

Basically what this button does is it will delete all webview data generated by the extensions trying to download/load chapters.
One downside of this feature is that it will clear any logins and site preferences data inside webview folder.

### My Suggestion
To allow implement this feature but inside the extension settings to not have it affect all other extensions data. 

And this settings will either **auto clear webview data for that specific extension** after each error faced during page list fetching since that is when the error throws according to my debugging. 

If auto clearing is not possible maybe consider implementing it as an option to NOT save webview data for that extension or something similar to this. Or maybe save the extension webview data inside a separated folder inside the /webview folder in the application's data folder and let Tachiyomi app implement the clearing feature based on extension basis.

If this is possible this would help tremendously with cloudflare bypass detection and loops issues since they mainly occur on a specific extensions and is not considered a common issue with most of the sources. This will help a large user base especially non-English users who are forced to use low quality sources due to lack of quality sources.

arkon:
Closes #940.

### DIFF
--- a/src/help/guides/troubleshooting.md
+++ b/src/help/guides/troubleshooting.md
@@ -139,9 +139,9 @@ If the site is not reachable or has issues, all you can do is wait for the site 
 
 ## Solving Cloudflare issues
 
-**Note that WebView is not the same as using your browser.**
+### Loading the website in WebView
 
-You must solve the **Cloudflare** issue in **WebView**.
+Try to load the website in **WebView**. **Note that WebView is not the same as using your browser.**
 
 ::: guide How to open WebView
 1. Go to <NavigationText item="browse"/> in the bottom navbar.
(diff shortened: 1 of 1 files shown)


arkon:
https://deploy-preview-952--tachiyomi.netlify.app/help/guides/troubleshooting/#solving-cloudflare-issues

PR #960, opened 2022-10-27
Update home screenshots to Material You design

### LINKED ISSUE #933: Update Screenshots in Homepage
### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

### Steps to reproduce

-

### Expected behavior

-

### Actual behavior

-

### Other details

As per arkon, screenshots in the Home page should be updated to show Material You updates.

mezzode:
Resolves #933


arkon:
We actually need to use copyright-free content for those screenshots.

mezzode:
Ah makes sense. Do we also want to avoid using Mangadex as well, and if so is there an existing provider that's fine to use? I'd prefer not to have to set up a whole Komga library just for this

mezzode:
I went ahead and switched to using more of Shuho Sato's work since it seems we were comfortable using that before

arkon:
Seishirou#1283 on Discord actually has a Komga set up, but that works too.

That said, are you doing the screenshots based on v0.13.6? There's some minor changes coming up in the next major release, so it may be worth doing it against the preview version.

alessandrojean:
Maybe compress the images and also make them the same width and height from the current ones? They seem to be too larger in width/height and file size too.

mezzode:
> That said, are you doing the screenshots based on v0.13.6? There's some minor changes coming up in the next major release, so it may be worth doing it against the preview version.

These were from v0.13.6. I just looked at the preview version and I'm not seeing any noticeable differences tho?

> Maybe compress the images and also make them the same width and height from the current ones? They seem to be too larger in width/height and file size too.

Will fix, forgot to do this when I switched to the copyright-free pics

arkon:
> I just looked at the preview version and I'm not seeing any noticeable differences tho?

The only obvious difference are the action icons in the manga view's toolbar. Otherwise, it'd be very minor padding changes or something, if anything.

We can also just leave the reader image the same, since there'd be no difference.

mezzode:
Updated the pics

Soitora:
> Updated the pics

Smh now u added back the tags used when multiple lang selected for souce
image

Not that it matters a ton

PR #991, opened 2023-01-07
add markdown linting

AriaMoradi:
A lot of files still need line breaking. I think that should be dealt with in time.

### DIFF
--- a/src/sandbox/style-guide.md
+++ b/src/sandbox/style-guide.md
@@ -7,15 +7,18 @@ sitemap:
 ---
 
 # Style guide
+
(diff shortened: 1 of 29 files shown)


src/sandbox/style-guide.md
```diff
@@ -77,6 +91,8 @@ Some **DO**s and **DO NOT**s for [FAQ](../help/faq/README.md).
 > ::::
 
 - **MAYBE** use accordions for FAQ entries, needs further testing if/before we replace all:
+
+exmaple 1:
```
arkon, line 95: "example" 

Same elsewhere.

arkon (changes requested):

PR #1008, opened 2023-04-15
Update reader settings guide

scb261:
### Merge only after next stable release

Based on r5675 and https://github.com/tachiyomiorg/tachiyomi/commit/8a18e10cc2b0fb3ed03bb05c3a53ceb6867d0a75

Todo:
- [ ] Update `Invert tap zones`. Currently the explanation is based on `Right and Left` tap zones which were the only possible tap zones back then. If anyone wants to do it - feel free
- [ ] Rename `Pan wide images when tapping` https://github.com/tachiyomiorg/tachiyomi/issues/9343
- [ ] Check if the beta feature is renamed/included in the stable
- [ ] Generally check if everything in the PR is still up-to-date after the next stable release

What was changed:
- Updated outdated info
- Added explanation for missing settings
- Added punctuations `.` and **bold text** in some places
- Removed tip from scale type because it actually applies to a lot of those settings
- Added `Paged` and `Long Strip` badges to settings that are present in both of those sections to highlight that you need to change them separately

### DIFF
--- a/src/help/guides/reader-settings.md
+++ b/src/help/guides/reader-settings.md
@@ -7,33 +7,36 @@ lang: en-US
 # Reader settings
 This section relates to the reading experience in the app and navigating the reader.
 
-## Default reading mode <Badge text="Right to left" type="default-indicator" vertical="middle" />
+## Default reading mode <Badge text="Paged (right to left)" type="default-indicator" vertical="middle" />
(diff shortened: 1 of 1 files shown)


Soitora:
Could still be relevant for rewrite site

PR #1024, opened 2023-07-30
Attempt to fix ESlint Errors on Netlify

Soitora:
Move `npm run lint` to outside of the `npm run build` script.
Fix ESLint errors Local, same error showing in Netlify console.
Add a new `npm run test` script, replacing the old "lint and build"

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,7 +35,7 @@ $ npm run build
 $ npm run lint
 ```
 
- **Please make sure to have `npm run build` pass successfully before submitting a PR.** Although the same tests will be run against your PR on the CI server, it is better to have it working locally.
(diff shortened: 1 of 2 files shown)


Soitora:
First in a while to succeed

image

arkon:
I'm not sure just skipping it really qualifies as fixing?

arkon:
Superseded by #1025

Soitora:
> I'm not sure just skipping it really qualifies as fixing?

The fix was actually the quote change, not the "while I am at it" script changes

arkon:
The quote changes aren't actually relevant, the linting just never runs if you check the deployment build logs.

Regardless, you just need to use a newer version of Node.

PR #1034, opened 2023-08-27
Optimize images to WebP

Soitora:
Changes `png`/`jpg` media to `webp`, changing overall asset size from `11,2 MB` to `1,93 MB`, an 83% reduction.

Logo in the root folder is kept as `png`, if I recall correctly, that specifically is best kept as `png` for compatibility reasons, which the other media don't suffer from.

Done with the help of [flozz/yoga-image-optimizer](https://github.com/flozz/yoga-image-optimizer), which uses [Zopflipng](https://github.com/google/zopfli) for PNG and [libwebp](https://chromium.googlesource.com/webm/libwebp/) for WEBP


arkon:
> compatibility reasons

With what?

Soitora:
> > compatibility reasons
> 
> With what?

I've seen some services that output website previews dislike `webp`, but that was some years ago, might not be relevant anymore?

PR #1113, opened 2024-01-09
Add extension creation guide

Soitora:
Adopted from https://github.com/tachiyomiorg/extensions/blob/main/CONTRIBUTING.md


arkon:
A lot of this is actually specific to how our repo in particular was set up. I'll take a pass at some point to make it more generic about what the expected outputs actually are and not necessarily how to copy how the relatively janky way that we were doing it was.

